PR #4, opened 2020-03-08
Add fish integration

alin23:
This is a very useful tool! Thanks for creating it and making it open source!

I added support for the fish shell in case anyone is interested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@
 
 A cd command that learns your habits
 
+## Table of contents
+
+- [Installing `zoxide`](#installing-zoxide)
+- [Adding `zoxide` to your shell](#adding-zoxide-to-your-shell)
+    + [zsh](#zsh)
+    + [fish](#fish)
+
 ## Introduction
 
 `zoxide` is a new `cd` alternative inspired by [`z`](https://github.com/rupa/z) and [`z.lua`](https://github.com/skywind3000/z.lua). It keeps track of the directories you use most frequently, and uses a ranking algorithm to navigate to the best match.
@@ -76,6 +83,12 @@ alias zi="z -i"
 alias za="zoxide add"
 alias zq="zoxide query"
 alias zr="zoxide remove"
+```
+
+#### fish
+Using [fisher](https://github.com/jorgebucaran/fisher):
+```sh
+fisher add ajeetdsouza/zoxide
 ```
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -4,13 +4,6 @@
 
 A cd command that learns your habits
 
-## Table of contents
-
-- [Installing `zoxide`](#installing-zoxide)
-- [Adding `zoxide` to your shell](#adding-zoxide-to-your-shell)
-    + [zsh](#zsh)
-    + [fish](#fish)
-
 ## Introduction
 
 `zoxide` is a new `cd` alternative inspired by [`z`](https://github.com/rupa/z) and [`z.lua`](https://github.com/skywind3000/z.lua). It keeps track of the directories you use most frequently, and uses a ranking algorithm to navigate to the best match.
@@ -90,6 +83,20 @@ Using [fisher](https://github.com/jorgebucaran/fisher):
 ```sh
 fisher add ajeetdsouza/zoxide
 ```
+
+Using [omf](https://github.com/oh-my-fish/oh-my-fish):
+```sh
+omf install https://github.com/ajeetdsouza/zoxide
+```
+
+To enable key bindings you have to set the universal variable `ZOXIDE_KEY_BINDINGS` to 1 and restart the shell. 
+```fish
+set -U ZOXIDE_KEY_BINDINGS 1
+```
+
+| Keybinding | Action                                         |
+| ---------- | -------------------------------------------------------------------------------- |
+| <kbd>Alt</kbd>+<kbd>z</kbd> / <kbd>⌥</kbd>+<kbd>z</kbd> | Launch Zoxide in interactive mode |
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Using [fisher](https://github.com/jorgebucaran/fisher):
 fisher add ajeetdsouza/zoxide
 ```
 
-Using [omf](https://github.com/oh-my-fish/oh-my-fish):
+Using [oh-my-fish](https://github.com/oh-my-fish/oh-my-fish):
 ```sh
 omf install https://github.com/ajeetdsouza/zoxide
 ```
--- a/README.md
+++ b/README.md
@@ -89,15 +89,6 @@ Using [omf](https://github.com/oh-my-fish/oh-my-fish):
 omf install https://github.com/ajeetdsouza/zoxide
 ```
 
-To enable key bindings you have to set the universal variable `ZOXIDE_KEY_BINDINGS` to 1 and restart the shell. 
-```fish
-set -U ZOXIDE_KEY_BINDINGS 1
-```
-
-| Keybinding | Action                                         |
-| ---------- | -------------------------------------------------------------------------------- |
-| <kbd>Alt</kbd>+<kbd>z</kbd> / <kbd>⌥</kbd>+<kbd>z</kbd> | Launch Zoxide in interactive mode |
-
 ## Configuration
 
 ### Environment variables
--- a/functions/z.fish
+++ b/functions/z.fish
@@ -1,0 +1,12 @@
+function z
+    if test (count $argv) -gt 0
+        set _Z_RESULT (zoxide query $argv)
+        switch "$_Z_RESULT"
+            case 'query: *'
+                cd (string sub -s 8 -- "$_Z_RESULT")
+            case '*'
+                echo "$_Z_RESULT"
+        end
+        commandline -f repaint
+    end
+end
--- a/functions/z.fish
+++ b/functions/z.fish
@@ -4,9 +4,12 @@ function z
         switch "$_Z_RESULT"
             case 'query: *'
                 cd (string sub -s 8 -- "$_Z_RESULT")
+                commandline -f repaint
             case '*'
-                echo "$_Z_RESULT"
+                if test -n "$_Z_RESULT"
+                    echo "$_Z_RESULT"
+                end
         end
-        commandline -f repaint
     end
 end
+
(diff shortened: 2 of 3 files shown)


init.fish
```diff
@@ -0,0 +1,14 @@
...
+    end
+end
+
+if command -q zoxide
+    abbr -a zi "z -i"
+    abbr -a za "zoxide add"
+    abbr -a zq "zoxide query"
+    abbr -a zr "zoxide remove"
+
+    bind \ez 'z -i'
```
ajeetdsouza: While this is definitely a useful keybinding, it would be inconsistent across shells. I think we should let users create their own customizations, rather than have a preinstalled keybinding.

README.md
```diff
@@ -4,6 +4,13 @@
 
 A cd command that learns your habits
 
+## Table of contents
```
ajeetdsouza: This is a great addition to the README. Could you create a separate PR for this?

README.md
```diff
@@ -78,6 +85,12 @@ alias zq="zoxide query"
 alias zr="zoxide remove"
 ```
 
+#### fish
+Using [fisher](https://github.com/jorgebucaran/fisher):
```
ajeetdsouza: Can we make this generic to all fish plugin managers?

ajeetdsouza:
I'm not too familiar with the fish plugin format - can we add `z` to the `init.fish` file instead of the `functions` folder? Creating a separate folder with a name like `functions` in the repository can be confusing.

Normally, there should have been another repo for the fish plugin, but I'm trying to keep things simple for the time being.

init.fish
```diff
@@ -0,0 +1,14 @@
+function zoxide-add --on-event fish_prompt
+    if command -q zoxide
```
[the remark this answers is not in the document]
alin23: In fish 3.1 `command -q` implies `-s` automatically, but I didn't know fish 3.0 doesn't also do this. It is a very inexpensive operation, but if you'd rather throw an error and performance is not the concern here, I will remove it.

functions/z.fish
```diff
@@ -0,0 +1,12 @@
+function z
+    if test (count $argv) -gt 0
+        set _Z_RESULT (zoxide query $argv)
+        switch "$_Z_RESULT"
+            case 'query: *'
+                cd (string sub -s 8 -- "$_Z_RESULT")
+            case '*'
+                echo "$_Z_RESULT"
+        end
+        commandline -f repaint
```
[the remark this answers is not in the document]
alin23: Actually `repaint` is only necessary if we cd as that changes the prompt. I'll fix that

init.fish
alin23: Ok, makes sense. Another way of doing this in fish world is to have a universal variable like `$ZOXIDE_KEY_BINDINGS` set to 0 by default. If the user sets it to 1 (`set -U ZOXIDE_KEY_BINDINGS 1) the key binding is turned on.

What do you think would be the best approach here?

README.md
alin23: Sure!

README.md
alin23: I'll also add support for `oh-my-fish`

alin23:
> I'm not too familiar with the fish plugin format - can we add `z` to the `init.fish` file instead of the `functions` folder? Creating a separate folder with a name like `functions` in the repository can be confusing.
> 
> Normally, there should have been another repo for the fish plugin, but I'm trying to keep things simple for the time being.

Unfortunately the functions folder has to exist as that is how fish works. The init file is linked into the `~/.config/fish/conf.d` directory and loaded every time the shell is started. The files inside `functions` are loaded on-demand when you call any function that isn't defined to avoid shell init overhead. So the first time you would call `z`, fish would look for a file named `z.fish` inside `~/.config/fish/functions` and source it.

I say it is simpler to have the shell files in the same repo as it is more intuitive when someone wants to install a plugin without having to always take a look at the instructions.

init.fish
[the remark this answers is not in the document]
ajeetdsouza: I think throwing an error should be the way to go on this one. Silently ignoring the command if the `zoxide` binary isn't found can be very confusing to users.

init.fish
ajeetdsouza: Either way, it still results in an inconsistency across different shells. I think custom keybindings should be left upto the user for now.

alin23:
@ajeetdsouza I updated the code based on your review. Let me know if there's anything needed or if it is ready for merging.

init.fish
```diff
@@ -0,0 +1,8 @@
+function zoxide-add --on-event fish_prompt
```
cole-h: What is the benefit of using the `fish_prompt` event, as opposed to watching the `PWD` shell variable?

Ref: https://github.com/jethrokuan/z/blob/5e8d0e2c900d6a8c9704e153a8126f776d9d2f3c/conf.d/z.fish#L42

alin23: I don't think there's any benefit, It's just what makes sense to me. I want folders to be added to zoxide when I `cd` voluntarily, not when the `$PWD` is changed which can happen multiple times in a function without the prompt being updated.

cole-h: I've never used any functions that used `cd` in their internals, so I failed to take that into account. Fair enough. Thanks for your work on this!

alin23: It's not common for functions to do that, but it can happen so I thought I should make sure the zoxide database doesn't get unwanted entries. 😊

functions/z.fish
```diff
@@ -0,0 +1,15 @@
+function z
+    if test (count $argv) -gt 0
+        set _Z_RESULT (zoxide query $argv)
+        switch "$_Z_RESULT"
+            case 'query: *'
+                cd (string sub -s 8 -- "$_Z_RESULT")
+                commandline -f repaint
+            case '*'
+                if test -n "$_Z_RESULT"
```
ajeetdsouza: Instead of checking for a non-empty string, why not use `echo -n` instead? That way, if the string is empty, it won't output anything.

README.md
```diff
@@ -78,6 +78,17 @@ alias zq="zoxide query"
...
+```sh
+fisher add ajeetdsouza/zoxide
+```
+
+Using [omf](https://github.com/oh-my-fish/oh-my-fish):
```
ajeetdsouza: Could you change `omf` to `oh-my-fish` for clarity?

functions/z.fish
alin23: Yep, that works, thanks for the idea!

README.md
alin23: Done!

ajeetdsouza:
This is perfect, thank you so much for your contribution!

skorokithakis:
This may be an issue with my fisher, but using this plugin only works on the terminal where I installed it. If I exit and open a new terminal, `z` is unavailable until I run `fisher` again to install the plugin again.

alin23:
@skorokithakis you either need to update fish or fisher or both. Fish 3.0 introduced a conf.d directory that handles init scripts and the newest version of fisher knows to link the init.fish script from this repo into that directory so that it runs every time the shell is started.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

skorokithakis:
@alin23 I see, thank you. Can the script be made backwards-compatible? Fish is managed by my OS and I don't think I will be able to upgrade soon...

alin23:
The script is not the issue here. It is a fundamental change in how fish/fisher handle plugins. You can just copy and paste the `init.fish` contents in your `~/.config/fish/config.fish` file and copy `functions/z.fish` in your `~/.config/fish/functions/` directory.

This is what fisher used to do in older versions of fish. It's not a bug in fisher, it is just an incompatibility with your fish version.

cole-h:
Hmm, something I just came across... If you continuously press enter in a directory, its rank gets boosted every time (due to it triggering on prompt display).

IMHO, I think it would be better to track on `PWD` change, despite the possibility of adding "background `cd`'d" directories to the list. I don't necessarily want a directory that I accidentally hold enter in to become the highest ranked.

Thoughts, @alin23?

If you find this undesirable, maybe we could implement some way to specify when we want it to run -- `on-event` or `on-variable`.

alin23:
@cole-h indeed, that is not how we want this to work. We could use the `fish_postexec` event, which sends the executed command as `$argv[1]`

Something like this:
```fish
function init_z --on-event fish_postexec
    if contains cd (string split -- ' ' "$argv[1]")
        zoxide add
    end
end
```

Maybe you could test this more extensively and make a PR with the change if it works.

ajeetdsouza:
> If you continuously press enter in a directory, its rank gets boosted every time (due to it triggering on prompt display).

Currently, this is intended behaviour, and works consistently across shells. I'm considering adding an environment variable that would allow users to tweak this. Could you move further discussion to an issue, where it can be discussed properly?

cole-h:
> @cole-h indeed, that is not how we want this to work. We could use the `fish_postexec` event, which sends the executed command as `$argv[1]`
> 
> Something like this:
> 
> ```fish
> function init_z --on-event fish_postexec
>     if contains cd (string split -- ' ' "$argv[1]")
>         zoxide add
>     end
> end
> ```
> 
> Maybe you could test this more extensively and make a PR with the change if it works.

There are a few problems I see with that approach: 1) it can add the current directory again in case of a failed `cd` e.g. into a nonexistent directory; and 2) by relying on the presence of the first argument being `cd`, it won't add directories that are jumped to using this project (or even if `cd` is aliased to e.g. `c`). (IMHO) I think tracking `PWD` is the only real solution here.

@ajeetdsouza I'm currently drafting an issue for discussion on this topic.

alin23:
@cole-h this is what I use right now and it's precise enough. It doesn't account for aliases to the cd command though. That can be fixed by using a universal variable instead of the literal `cd`.
```fish
function __zoxide_check_pwd_changed --on-event fish_preexec
    if contains cd (string split -- ' ' "$argv[1]")
        set -g __CD_COMMAND_EXEC "$argv[1]"
        set -g __CD_COMMAND_PWD "$PWD"
    end
end

function __zoxide_add --on-event fish_postexec
    if set -q __CD_COMMAND_EXEC;
        and test "$__CD_COMMAND_EXEC" = "$argv[1]";
        and set -q __CD_COMMAND_PWD;
        and test "$__CD_COMMAND_PWD" != "$PWD"
        zoxide add
    end
end
```

I tested a PWD listener for a day and my zoxide database got filled with directories I never use because of some fish plugins that use cd inside. Turns out this does happen.

ajeetdsouza:
@skorokithakis, zoxide no longer requires fisher, I hope this fixes your issue!

skorokithakis:
@ajeetdsouza my issue turned out to be with fisher dying over a trailing slash in my env vars, I am now a happy zoxide user, thank you!

PR #10, opened 2020-03-09
fix(zsh): append `_zoxide_precmd` as a single-element array to hook

ErichDonGubler:
We should append `_zoxide_precmd` to `precmd_functions` within an array literal, not by itself.

This is important because...well, otherwise `precmd_functions` doesn't remain an array. One can only get an array assigned into the left-hand symbol by appending *another* array, because...well, that's what's expected in `zsh` and `bash`! Is it weird? Yes. But that's the type system we live with!


ajeetdsouza:
Didn't notice that. Thanks!

PR #19, opened 2020-03-09
Track Cargo.lock

cust0dian:
This [allows for reproducible builds](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) and makes it easier to package `zoxide`.


ajeetdsouza:
Thank you for your contribution!

Issue #8, opened 2020-03-08
Add an init subcommand


Lugoues:
Would be great to have an init method. This would allow us to source `$(eval zoxide init zsh)` and stay up to date if you make a script change in the future.

cole-h:
I actually really like this idea. It's something that the "competing" [pazi] does and would alleviate the concern re: the fish script ("Creating a separate folder with a name like functions in the repository can be confusing." https://github.com/ajeetdsouza/zoxide/pull/4#issuecomment-596216668).

Without thinking too deeply, I can't see many downsides to this approach (there are probably more than I'm thinking of, however). The upside is that there is no need for a plugin manager, because the user only needs to add one eval to their config, and everything is then on pace with the binary itself.

Overall, I'm a fan.

[pazi]: https://github.com/euank/pazi#how-do-i-install-pazi

ajeetdsouza:
I agree, this is a great idea! I will work on this today.

ifreund:
I hope this can be implemented without hard coding the default aliases of `z`, `zi`, `za`, etc.  I prefer to use `j` for mine since autojump was the first such program I used and because `j` is right under my index finger on the homerow.

cole-h:
I think the best way to do this is to check for environment variables, like [`z.lua` does](https://github.com/skywind3000/z.lua/blob/master/README.md#options). Alternatively (not really a fan of this one, but...): could use command line flags for each thing to change.

ajeetdsouza:
I've added an `init` subcommand in https://github.com/ajeetdsouza/zoxide/commit/f0c5e28fd7e56d7cec045a40154ab4768339da44. Currently, I still haven't added the environment variable, so I'll leave this issue open for now.

ajeetdsouza:
I've added a command line flag in https://github.com/ajeetdsouza/zoxide/commit/34ab8f3f8b502f11ba51dd69af75ee7f4dcf0654. I felt a flag would be better than an environment variable here since it only affects this particular command and not the entire program.

PR #21, opened 2020-03-10
add bash support; installation instructions

NerdyPepper:
- requires installation of the [`bash-preexec`](https://github.com/rcaloras/bash-preexec) script
- no support for any bash plugin managers (i have no experience with any of them)

i hope this pr aligns with the vision of this project, i am willing to make changes as you see fit. thank you for your work on `zoxide`!

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ A cd command that learns your habits
   - [Adding `zoxide` to your shell](#adding-zoxide-to-your-shell)
     - [zsh](#zsh)
     - [fish](#fish)
+    - [bash](#bash)
 - [Configuration](#configuration)
   - [Environment variables](#environment-variables)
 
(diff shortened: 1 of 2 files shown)


ajeetdsouza:
`bash-preexec` seems to be using `PROMPT_COMMAND` under the hood, why not do it ourselves and remove the dependency?

In line with https://github.com/ajeetdsouza/zoxide/issues/8, I've been trying to move away from the dependency on package managers, which is why the move to `bash` took so long. I've added it in https://github.com/ajeetdsouza/zoxide/commit/f0c5e28fd7e56d7cec045a40154ab4768339da44, so I'm closing this PR now.

All the same, thanks for your work, I really appreciate it!

NerdyPepper:
I am thrilled to see that commit land! Thanks and no problem!

PR #29, opened 2020-03-12
Migrate from `z` database

### LINKED ISSUE #11: Import the entries from .z
I started using zoxide and am very impressed by its speed compared to other utilities. Am currently using zsh-z. It would be nice if there was a way to import the entries from the earlier $HOME/.z to zoxide database. Any plans on using the existing data?

cole-h:
The new `migrate` function takes in a path to the old `z` database and
naively parses it to add to the database. It sets the epoch to be the
current time in order to prevent scaling down old directories from the
`z` database -- at least until `zoxide` is up and running.

The program will fail if the user already has a database, so as to
prevent tainting it in any way. It will also fail if the database is not
in the expected format -- e.g. if it doesn't follow the
`/path/to/somewhere|1.0101|10101010101` schema. The way this is
validated is by looking for 2 pipe characters and erroring if there is
only 1 (`first_pipe == last_pipe`). It will ignore any dead paths if the
user has not cleaned their database recently.

---

Fix #11 

Do note that, in its current implementation, only a database with `/path/to/somewhere|1.0101|10101010101` format is supported. If there are any other formats, those could be added down the line, as needed.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -79,7 +79,8 @@ impl DB {
             );
         }
 
-        let z_db_file = File::open(path).with_context(|| anyhow!("could not open z db file"))?;
+        let z_db_file =
+            File::open(path).with_context(|| anyhow!("could not open z database file"))?;
         let reader = BufReader::new(z_db_file);
 
         for (idx, read_line) in reader.lines().enumerate() {
@@ -95,16 +96,38 @@ impl DB {
 
             match split_line.as_slice() {
                 [epoch_str, rank_str, path_str] => {
-                    let epoch = epoch_str
-                        .parse::<i64>()
-                        .with_context(|| anyhow!("could not parse epoch: '{}'", epoch_str))?;
-                    let rank = rank_str
-                        .parse::<f64>()
-                        .with_context(|| anyhow!("could not parse rank: '{}'", rank_str))?;
-                    let path = match PathBuf::from(path_str).canonicalize() {
+                    let epoch = match epoch_str.parse::<i64>() {
+                        Ok(epoch) => epoch,
+                        Err(e) => {
+                            eprintln!(
+                                "invalid epoch '{}' at line {}: {}",
+                                epoch_str, line_number, e
+                            );
+                            continue;
+                        }
+                    };
+                    let rank = match rank_str.parse::<f64>() {
+                        Ok(rank) => rank,
+                        Err(e) => {
+                            eprintln!("invalid rank '{}' at line {}: {}", rank_str, line_number, e);
+                            continue;
+                        }
+                    };
+                    let path_abs = match Path::new(path_str).canonicalize() {
                         Ok(path) => path,
                         Err(e) => {
                             eprintln!("invalid path '{}' at line {}: {}", path_str, line_number, e);
+                            continue;
+                        }
+                    };
+                    let path_str = match path_abs.to_str() {
+                        Some(path) => path,
+                        None => {
+                            eprintln!(
+                                "invalid unicode in path '{}' at line {}",
+                                path_abs.display(),
+                                line_number
+                            );
                             continue;
                         }
                     };
@@ -112,7 +135,7 @@ impl DB {
                     // FIXME: When we switch to PathBuf for storing directories inside Dir, just
                     // pass `PathBuf::from(path_str)`
                     self.dirs.push(Dir {
-                        path: path.display().to_string(),
+                        path: path_str.to_string(),
                         rank,
                         last_accessed: epoch,
                     });
--- a/src/db.rs
+++ b/src/db.rs
@@ -73,53 +73,71 @@ impl DB {
 
     pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
         if !self.dirs.is_empty() {
+            // FIXME: just for testing
+            #[cfg(not(debug_assertions))]
             return Err(anyhow!(
                 "To prevent conflicts, you can only migrate from z with an empty \
                  zoxide database!"
             ));
         }
 
-        let zdata = File::open(path)?;
-        let reader = BufReader::new(zdata);
+        let z_db_file = File::open(path).with_context(|| anyhow!("could not open z db file"))?;
+        let reader = BufReader::new(z_db_file);
 
-        for line in reader.lines().filter_map(|l| l.ok()) {
-            let first_pipe_idx = line
-                .find('|')
-                .with_context(|| anyhow!("missing separator"))?;
-            let last_pipe_idx = line
-                .rfind('|')
-                .with_context(|| anyhow!("missing separator"))?;
-
-            if first_pipe_idx == last_pipe_idx {
-                return Err(anyhow!("invalid data file format -- only 1 separator"));
-            }
-
-            let path = PathBuf::from(&line[..first_pipe_idx]);
-            let rank = line[first_pipe_idx + 1..last_pipe_idx]
-                .parse::<f64>()
-                .with_context(|| anyhow!("could not parse rank"))?;
-
-            // otherwise, the rank will get scaled down, depending on how old
-            // the entry is
-            let epoch = util::get_current_time()?;
-
-            let path_abs = match path.canonicalize() {
-                Ok(path) => path,
-                Err(_) => continue, // ignore dead paths
+        for (idx, read_line) in reader.lines().enumerate() {
+            let line_number = idx + 1;
+            let line = if let Ok(line) = read_line {
+                line
+            } else {
+                eprintln!("could not read line {}: {:?}", line_number, read_line);
+                continue;
             };
 
-            let path_str = path_abs
-                .to_str()
-                .ok_or_else(|| anyhow!("invalid unicode in path: {}", path_abs.display()))?;
+            let split_line = line.rsplitn(3, '|').collect::<Vec<&str>>();
 
-            self.dirs.push(Dir {
-                path: path_str.to_owned(),
-                last_accessed: epoch,
-                rank,
-            });
+            match split_line.as_slice() {
+                [_, rank_str, path_str] => {
+                    let path = PathBuf::from(path_str);
+                    let rank = rank_str
+                        .parse::<f64>()
+                        .with_context(|| anyhow!("could not parse rank"))?;
+
+                    // otherwise, the rank will get scaled down, depending on
+                    // how old the entry is (import in-place and let zoxide
+                    // scale down organically)
+                    let epoch = util::get_current_time()?;
+
+                    // TODO: does this *need* to be canonicalized?
+                    // 1) the z db already stores canonicalized paths
+                    // 2) zoxide purges nonexistent paths upon query invocation
+                    let path_abs = match path.canonicalize() {
+                        Ok(path) => path,
+                        Err(_) => continue, // ignore dead paths
+                    };
+
+                    // FIXME: Remove when we switch to PathBuf for storing
+                    // diectories inside Dir and just pass `path` (the PathBuf
+                    // constructed from `path-str`)
+                    let path_str = path_abs.to_str().ok_or_else(|| {
+                        anyhow!("invalid unicode in path: {}", path_abs.display())
+                    })?;
+
+                    self.dirs.push(Dir {
+                        path: path_str.to_owned(),
+                        last_accessed: epoch,
+                        rank,
+                    });
+                }
+                [] | [""] => (),
+                line => {
+                    eprintln!("invalid line {}: {:?}", line_number, line);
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -71,6 +71,61 @@ impl DB {
...
+            ));
+        }
+
+        let zdata = File::open(path)?;
+        let reader = BufReader::new(zdata);
+
+        for line in reader.lines() {
+            let line = line?;
+
+            let first_pipe = line
```
lovesegfault: I think it's better if these are named in a way that makes clear they're just a position within `line`.

src/db.rs
```diff
@@ -71,6 +71,61 @@ impl DB {
...
+                "To prevent conflicts, you can only migrate from z with an empty \
+                 zoxide database!"
+            ));
+        }
+
+        let zdata = File::open(path)?;
+        let reader = BufReader::new(zdata);
+
+        for line in reader.lines() {
+            let line = line?;
```
lovesegfault: Instead of doing this consider writing a proper iterator chain using `filter_map` and the like

src/db.rs
```diff
@@ -71,6 +71,61 @@ impl DB {
...
+            // otherwise, the rank will get scaled down, depending on how old
+            // the entry is
+            let epoch = util::get_current_time()?;
+
+            let path_abs = match path.canonicalize() {
+                Ok(path) => path,
+                Err(_) => continue, // ignore dead paths
+            };
+
+            let path_str = path_abs
```
lovesegfault: You should not be using `String` to represent paths, use `OsString`

cole-h: (Covered this on IRC, but leaving it here for posterity/conversation's sake)

I agree. But I believe it is out of scope for this PR and should be done somewhere else (either a follow-up PR, a commit by our friendly neighborhood maintainer, or a PR prior to landing this).

src/db.rs
cole-h: Just so people know: I did attempt to `for_each` this, but 1) the `anyhow!` macro doesn't work inside of closures; and 2) you can't `return` from a function while inside a closure.

I did forget about `filter_map`, so I did at least include that.

src/db.rs
ajeetdsouza: I think `PathBuf` would be more fitting, and `serde` supports it too. I kept it as a `String` for the initial release for simplicity of string matching, but I will change this in a later commit.

src/db.rs
```diff
@@ -71,6 +71,59 @@ impl DB {
...
 
+    pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
+        if !self.dirs.is_empty() {
+            return Err(anyhow!(
+                "To prevent conflicts, you can only migrate from z with an empty \
+                 zoxide database!"
+            ));
+        }
+
+        let zdata = File::open(path)?;
```
ajeetdsouza: Add `with_context` here, so that the user can see what exactly caused the error. Also, a more descriptive name would be nice, like `z_file`.

src/db.rs
```diff
@@ -71,6 +71,59 @@ impl DB {
...
+        let reader = BufReader::new(zdata);
```
ajeetdsouza: I'm normally not a fan of short variables, but naming readers `rd` for brevity seems to be standard in Rust code. Could you rename this to `rd`?

src/db.rs
```diff
@@ -71,6 +71,59 @@ impl DB {
...
+
+        for line in reader.lines().filter_map(|l| l.ok()) {
```
ajeetdsouza: I think the line read error should be explicitly handled.

src/db.rs
```diff
@@ -71,6 +71,59 @@ impl DB {
...
+            let first_pipe_idx = line
```
ajeetdsouza: Rather than working with indices, how about the more concise:

```rust
for line_res in rd.lines() {
    let line = line_res.with_context(|| anyhow!("could not read line"))?;
    let line_split = line.rsplitn(3, '|').collect::<Vec<&str>>();

    match line_split.as_slice() {
        [time_str, rank_str, path_str] => todo!(),
        [] | [""] => (),
        _ => return Err(todo!()),
    };
}
```

src/db.rs
cole-h: `z_db` or `z_database`, maybe? `z_file` seems even less descriptive than `zdata`, IMHO.

ajeetdsouza:
Additionally, what is your take on doing a "best effort" import? Instead of throwing parser errors, we could simply raise a warning to the user and ignore that line. 

Of course, any `io::Error` should be thrown immediately.

src/db.rs
cole-h: Honestly, I'm not a fan of renaming it to `rd` -- solely because it goes against what you asked me above when opening the file ("a more descriptive name"). I can't claim to have read a lot of Rust codebases, but I haven't ever seen a reader named `rd`.

Unless you very strongly insist, I think I'll refrain from making this change.

src/db.rs
cole-h: Any reason why you chose `rsplitn` over `splitn`?

ajeetdsouza: The path name might have a `|` in it, but the timestamp and rank won't. Using `rsplitn` will take the `timestamp` and `rank` out first, and then use the rest of the line as the `path`.

cole-h:
I'd be fine with a "best effort" import. How would you prefer these warnings displayed? `eprintln!`?

src/db.rs
ajeetdsouza: I did notice the irony, but I've seen something `fn foo<R: Read>(rd: R)` in a lot of places.

However, I just checked https://github.com/rust-lang/rust, and it seems they're using `reader` for their `BufReader`s. So, you win!

src/db.rs
cole-h: Maybe we could meet in the middle and do `z_dbfile` or `z_db_file`? ;)

ajeetdsouza: `z_db_file` sounds fine!

ajeetdsouza:
Using a logging crate seems like overkill. For this PR, I'd say use `eprintln!`, but if you feel a logging crate would be better design, do open an issue where we can explore migration to different logging crates.

cole-h:
I agree that logging apparatus is too much. `eprintln!` it is!

src/db.rs
```diff
@@ -71,6 +71,77 @@ impl DB {
...
+                    let rank = rank_str
+                        .parse::<f64>()
+                        .with_context(|| anyhow!("could not parse rank"))?;
+
+                    // otherwise, the rank will get scaled down, depending on
+                    // how old the entry is (import in-place and let zoxide
+                    // scale down organically)
+                    let epoch = util::get_current_time()?;
+
+                    // TODO: does this *need* to be canonicalized?
```
cole-h: @ajeetdsouza What is your opinion on this?

The `z` database already stores `canonicalize`d paths, so I think it would make sense to remove `path_abs` and `path_str` below. As it stands, I have a little bit of duplicated code: `path_str` is transformed into a path (`PathBuf::from(path_str)`), which is canonicalized, which is then transformed into a string again (`let path_str = path_abs.to_str().....`).

src/db.rs
```diff
@@ -71,6 +71,77 @@ impl DB {
...
+
+        let z_db_file = File::open(path).with_context(|| anyhow!("could not open z db file"))?;
+        let reader = BufReader::new(z_db_file);
+
+        for (idx, read_line) in reader.lines().enumerate() {
+            let line_number = idx + 1;
+            let line = if let Ok(line) = read_line {
+                line
+            } else {
+                eprintln!("could not read line {}: {:?}", line_number, read_line);
```
cole-h: My "best effort" at "best effort" parsing. Might have missed some things; do let me know if that happens to be the case.

src/db.rs
cole-h: I decided to rely on the fact that the paths inside `z`'s database are already canonicalized. Let me know if you disagree with this course of action and I'd be happy to talk about it.

ajeetdsouza: I don't think we should make assumptions about the kind of data we're getting, especially since we're importing a format that a lot of different applications use (`z`, `z.lua`, `zsh-z`, etc.). While a rogue entry wouldn't really do much harm, an unresolved symbolic link would remain in the database for a while before its rank finally drops to < 1.

src/db.rs
```diff
@@ -71,6 +71,59 @@ impl DB {
...
+                continue;
+            };
+
+            let split_line = line.rsplitn(3, '|').collect::<Vec<&str>>();
+
+            match split_line.as_slice() {
+                [_, rank_str, path_str] => {
+                    // Set last_accessed to current time; otherwise, the rank will get scaled down,
```
ajeetdsouza: I don't think we should do this. IMO, the frecency _should_ get scaled down if the entry is old. Immediately after importing, if all the entries are old, all of them will be scaled equally, and there will be no problem. Once the user starts using `zoxide`, the new entries created will be prioritized, which is also intended behavior.

src/db.rs
```diff
@@ -71,6 +71,59 @@ impl DB {
         Ok(())
     }
 
+    pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
+        if !self.dirs.is_empty() {
+            return Err(anyhow!(
```
ajeetdsouza: Import and use the `anyhow::bail` macro here.

src/db.rs
cole-h: You're right -- the smaller the amount of assumptions, the less of a chance of things going wrong.

src/db.rs
cole-h: Fair.

src/db.rs
```diff
@@ -94,26 +94,32 @@ impl DB {
...
+                            eprintln!("invalid path '{}' at line {}: {}", path_str, line_number, e);
+                            continue;
+                        }
+                    };
 
                     // FIXME: When we switch to PathBuf for storing directories inside Dir, just
                     // pass `PathBuf::from(path_str)`
                     self.dirs.push(Dir {
-                        path: path_str.to_string(),
+                        path: path.display().to_string(),
```
ajeetdsouza: AFAIK, this is a lossy conversion, and would create an invalid database entry. Use `to_str()` instead, and warn the user if the path contains invalid Unicode.

src/db.rs
```diff
@@ -94,26 +94,32 @@ impl DB {
...
-                    // scale down organically)
-                    let epoch = util::get_current_time()?;
+                [epoch_str, rank_str, path_str] => {
+                    let epoch = epoch_str
+                        .parse::<i64>()
+                        .with_context(|| anyhow!("could not parse epoch: '{}'", epoch_str))?;
                     let rank = rank_str
                         .parse::<f64>()
                         .with_context(|| anyhow!("could not parse rank: '{}'", rank_str))?;
+                    let path = match PathBuf::from(path_str).canonicalize() {
```
ajeetdsouza: Use `Path::new`, since we don't have an owned `String`.

src/db.rs
```diff
@@ -71,6 +71,65 @@ impl DB {
...
+                continue;
+            };
+
+            let split_line = line.rsplitn(3, '|').collect::<Vec<&str>>();
+
+            match split_line.as_slice() {
+                [epoch_str, rank_str, path_str] => {
+                    let epoch = epoch_str
+                        .parse::<i64>()
+                        .with_context(|| anyhow!("could not parse epoch: '{}'", epoch_str))?;
```
cole-h: @ajeetdsouza Since we're doing "best effort", should I `eprintln!` these too, instead of erroring? I think I should, but just wanted to make sure.

ajeetdsouza: Yes, avoid throwing errors unless they are related to I/O.

src/db.rs
cole-h: You are indeed correct -- I always forget about this ^^;

src/db.rs
```diff
@@ -71,6 +71,65 @@ impl DB {
...
 
+    pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
+        if !self.dirs.is_empty() {
+            bail!(
+                "To prevent conflicts, you can only migrate from z with an empty \
+                 zoxide database!"
+            );
+        }
+
+        let z_db_file = File::open(path).with_context(|| anyhow!("could not open z db file"))?;
```
ajeetdsouza: Change the error to `"could not open z database file"` for consistency.

ajeetdsouza:
@cole-h thank you for the fantastic job on this PR!

cole-h:
Thanks for bearing with me! Double thanks for squash-merging!

PR #33, opened 2020-03-13
Allow merging when migrating from `z`

cole-h:
If the user passes the `--merge` flag to the `migrate` subcommand, all
duplicate entries will have their ranks and epochs updated: the rank
will be the sum of the stored rank and the newly-parsed rank, while the
epoch will be the maximum of stored epoch and the newly-parsed epoch.

This allows one to import from the `z` database even after having used
zoxide for any amount of time. This also permits a user who has already
sourced the init script to import their old database without needing to
do something like `rm ~/.zo && zoxide migrate ~/.z`.

---

Inspired by https://github.com/ajeetdsouza/zoxide/issues/11#issuecomment-598916309.

One decision that needs to be made is: do we want to truncate the existing database, or just remove any conflicting paths before re-adding them with the migrated data? I opted for the latter at this time (less destructive seems nice), but I can see an argument for either way.

Thoughs, @ajeetdsouza?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -70,11 +70,11 @@ impl DB {
         Ok(())
     }
 
-    pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
-        if !self.dirs.is_empty() {
+    pub fn migrate<P: AsRef<Path>>(&mut self, path: P, merge: bool) -> Result<()> {
+        if !self.dirs.is_empty() && !merge {
             bail!(
-                "To prevent conflicts, you can only migrate from z with an empty \
-                 zoxide database!"
+                "To prevent conflicts, you can only migrate from z with an empty zoxide database!
+If you wish to merge the two, specify the `--merge` flag."
             );
         }
 
@@ -130,6 +130,17 @@ impl DB {
                             continue;
                         }
                     };
+
+                    if merge {
+                        // If the path exists in the database, add the ranks and set the epoch to
+                        // the largest of the parsed epoch and the already present epoch.
+                        if let Some(dir) = self.dirs.iter_mut().find(|dir| dir.path == path_str) {
+                            dir.rank += rank;
+                            dir.last_accessed = Epoch::max(epoch, dir.last_accessed);
+
+                            continue;
+                        };
+                    }
 
                     // FIXME: When we switch to PathBuf for storing directories inside Dir, just
                     // pass `PathBuf::from(path_str)`
--- a/src/db.rs
+++ b/src/db.rs
@@ -87,7 +87,10 @@ If you wish to merge the two, specify the `--merge` flag."
             let line = if let Ok(line) = read_line {
                 line
             } else {
-                eprintln!("could not read line {}: {:?}", line_number, read_line);
+                eprintln!(
+                    "could not read entry at line {}: {:?}",
+                    line_number, read_line
+                );
                 continue;
             };
 
--- a/src/subcommand/migrate.rs
+++ b/src/subcommand/migrate.rs
@@ -7,10 +7,13 @@ use structopt::StructOpt;
 #[structopt(about = "Migrate from z database")]
 pub struct Migrate {
     path: String,
+
+    #[structopt(long, help = "Merge entries into existing database")]
+    merge: bool,
 }
 
 impl Migrate {
     pub fn run(&self, env: &Env) -> Result<()> {
-        util::get_db(env)?.migrate(&self.path)
+        util::get_db(env)?.migrate(&self.path, self.merge)
     }
 }


ajeetdsouza:
Rather than `--force`, how about `--merge`?

For each common path,
- Add the `rank`
- Pick the greater of the 2 `timestamp`s

src/db.rs
```diff
@@ -70,11 +70,12 @@ impl DB {
...
     }
 
-    pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
-        if !self.dirs.is_empty() {
+    pub fn migrate<P: AsRef<Path>>(&mut self, path: P, merge: bool) -> Result<()> {
+        if !self.dirs.is_empty() && !merge {
             bail!(
                 "To prevent conflicts, you can only migrate from z with an empty \
-                 zoxide database!"
+                 zoxide database!\n\
```
cole-h: I wish this could look nicer, but we don't want overlong lines, especially in a terminal application.

cole-h:
I think this should be good to go, aside from my singular concern above.

src/db.rs
```diff
@@ -131,6 +132,17 @@ impl DB {
                         }
                     };
 
+                    if merge {
+                        // If the path exists in the database, add the ranks and set the epoch to
+                        // the largest of the parsed epoch and the already present epoch.
+                        if let Some(dir) = self.dirs.iter_mut().find(|dir| dir.path == path_str) {
+                            dir.rank += rank;
+                            dir.last_accessed = i64::max(epoch, dir.last_accessed);
```
ajeetdsouza: Use `Epoch::max` from `crate::types` instead.

src/db.rs
ajeetdsouza: [Rust does support multiline strings](https://www.programming-idioms.org/idiom/48/multi-line-string-literal/435/rust), so there isn't any need for the `\` and `\n`.

src/db.rs
```diff
@@ -131,6 +132,17 @@ impl DB {
...
+
+                            continue;
```
ajeetdsouza: I feel it would improve code clarity if there was an `else` clause rather than a `continue`.

cole-h: The only reason I didn't do that was to reduce duplicate code -- otherwise there would be two instances of `self.dirs.push(Dir {`: one for when `merge` is active and the path wasn't already stored, and one for when `merge` is inactive.

Maybe I could add a comment above the `continue`, explaining why? Or, if you insist, I could add an `else` block.

src/db.rs
cole-h: Indeed it does. But then you get code that looks like:

```rust
            bail!(
                "To prevent conflicts, you can only migrate from z with an empty \
                 zoxide database!
If you wish to merge the two, specify the `--merge` flag."
            );
```

which I'm not sure is much better.

Another option would be to use `format!("{}\n{}", "first", "second")`. Let me know if your opinion is unchanged and you would still prefer the multiline string above.

src/db.rs
cole-h: For reference, here's what would need to be done (unless I overlooked something, which is exceedingly likely):

<details><summary>diff</summary>

```diff
diff --git a/src/db.rs b/src/db.rs
index 7041b4f..c3e8c6f 100644
--- a/src/db.rs
+++ b/src/db.rs
@@ -135,21 +135,30 @@ impl DB {
                     if merge {
                         // If the path exists in the database, add the ranks and set the epoch to
                         // the largest of the parsed epoch and the already present epoch.
-                        if let Some(dir) = self.dirs.iter_mut().find(|dir| dir.path == path_str) {
-                            dir.rank += rank;
-                            dir.last_accessed = Epoch::max(epoch, dir.last_accessed);
-
-                            continue;
-                        };
+                        match self.dirs.iter_mut().find(|dir| dir.path == path_str) {
+                            Some(dir) => {
+                                dir.rank += rank;
+                                dir.last_accessed = Epoch::max(epoch, dir.last_accessed);
+                            }
+                            None => {
+                                // FIXME: When we switch to PathBuf for storing directories inside Dir, just
+                                // pass `PathBuf::from(path_str)`
+                                self.dirs.push(Dir {
+                                    path: path_str.to_string(),
+                                    rank,
+                                    last_accessed: epoch,
+                                });
+                            }
+                        }
+                    } else {
+                        // FIXME: When we switch to PathBuf for storing directories inside Dir, just
+                        // pass `PathBuf::from(path_str)`
+                        self.dirs.push(Dir {
+                            path: path_str.to_string(),
+                            rank,
+                            last_accessed: epoch,
+                        });
                     }
-
-                    // FIXME: When we switch to PathBuf for storing directories inside Dir, just
-                    // pass `PathBuf::from(path_str)`
-                    self.dirs.push(Dir {
-                        path: path_str.to_string(),
-                        rank,
-                        last_accessed: epoch,
-                    });
                 }
                 [] | [""] => {} // ignore blank lines
                 line => {
```
</details>

Note the two instances of `self.dirs.push(Dir {` now. We need to handle the case where the directory doesn't exist in the database, which was originally done by just letting the `if let Some(...` fall through.

src/db.rs
```diff
@@ -70,12 +70,13 @@ impl DB {
...
 
-    pub fn migrate<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
-        if !self.dirs.is_empty() {
-            bail!(
-                "To prevent conflicts, you can only migrate from z with an empty \
-                 zoxide database!"
-            );
+    pub fn migrate<P: AsRef<Path>>(&mut self, path: P, merge: bool) -> Result<()> {
+        if !self.dirs.is_empty() && !merge {
+            bail!(format!(
```
ajeetdsouza: For future reference, the `bail!` macro also does formatting, so the `format!` there is redundant. What I meant, though, is that something like this would be nice:


```rust
            bail!(
                "To prevent conflicts, you can only migrate from z with an empty zoxide database!
If you wish to merge the two, specify the `--merge` flag."
            );
```

src/db.rs
ajeetdsouza: You're right, I missed that.

src/subcommand/migrate.rs
```diff
@@ -7,10 +7,13 @@ use structopt::StructOpt;
 #[structopt(about = "Migrate from z database")]
 pub struct Migrate {
     path: String,
+
+    #[structopt(long, help = "Merge conflicting entries")]
```
ajeetdsouza: I think a help message of "Merge entries into existing database" would be more accurate.

src/db.rs
cole-h: I personally am not a fan of multiline strings that de-indent, but I do agree that it looks nicer than both a `format!` (or just `"{}\n{}"`) and `\n\`.

src/subcommand/migrate.rs
cole-h: Agreed. Thanks!

cole-h:
I snuck a second commit in here to bring the message displayed when a line fails to parse in-line with what I added in the initial implementation. It's just a small issue and can be easily dropped if you don't see a need for it.

ajeetdsouza:
Thank you for your work on this!

PR #38, opened 2020-03-17
add armv7 release step

qmx:
This PR provides an armv7 build for zoxide - good for running it on devices like the raspberry pi 4


ajeetdsouza:
Added, thanks!

PR #43, opened 2020-03-18
Add POSIX shell support

### LINKED ISSUE #42: POSIX sh support
Would be nice to have POSIX sh support instead of just Bash. Would allow people to use zoxide on a wide variety of shells.

The Bash init script doesn't seem too far from being POSIX.

cole-h:
Any users of a shell that adheres to the POSIX standard should now be
supported. Shells that were tested while this feature was in development
include `mrsh`, `dash`, busybox `ash`, and `bash --posix`.

The hooks work by defining a `_zoxide_hook` function and adding it to
the shell's `PS1` (causing it to be evaluated every time the prompt is
redrawn).

The PWD hook uses a hack to get it to work properly. Variables are not
available outside of the functions defining them (even if `export`ed),
so we write the current PWD to a temporary file, `/tmp/.zo_pwd`. This is
not ideal, but I cannot think of a better solution apart from removing
the hook entirely.

---

Fix #42

### DIFF
--- a/src/subcommand/init.rs
+++ b/src/subcommand/init.rs
@@ -38,21 +38,20 @@ impl Init {
         let stdout = io::stdout();
         let mut handle = stdout.lock();
 
-        writeln!(handle, "{}", config.z)?;
+        // If any `writeln!` call fails to write to stdout, we assume the user's
+        // computer is on fire and panic.
+        writeln!(handle, "{}", config.z).unwrap();
         if !self.no_define_aliases {
-            writeln!(handle, "{}", config.alias)?;
+            writeln!(handle, "{}", config.alias).unwrap();
         }
 
         match self.hook {
             Hook::none => (),
-            Hook::prompt => writeln!(handle, "{}", config.hook.prompt)?,
-            Hook::pwd => {
-                if let Some(pwd_hook) = config.hook.pwd {
-                    writeln!(handle, "{}", pwd_hook)?;
-                } else {
-                    bail!("PWD hooks are currently unsupported on this shell.");
-                }
-            }
+            Hook::prompt => writeln!(handle, "{}", config.hook.prompt).unwrap(),
+            Hook::pwd => match config.hook.pwd {
+                Some(pwd_hook) => writeln!(handle, "{}", pwd_hook).unwrap(),
+                None => bail!("PWD hooks are currently unsupported on this shell."),
+            },
         }
 
         Ok(())
--- a/src/subcommand/init.rs
+++ b/src/subcommand/init.rs
@@ -1,6 +1,8 @@
+use anyhow::{bail, Result};
 use clap::arg_enum;
-use std::io::{self, Write};
 use structopt::StructOpt;
+
+use std::io::{self, Write};
 
 #[derive(Debug, StructOpt)]
 #[structopt(about = "Generates shell configuration")]
@@ -25,7 +27,7 @@ pub struct Init {
 }
 
 impl Init {
-    pub fn run(&self) {
+    pub fn run(&self) -> Result<()> {
         let config = match self.shell {
             Shell::bash => BASH_CONFIG,
             Shell::fish => FISH_CONFIG,
@@ -36,20 +38,24 @@ impl Init {
         let stdout = io::stdout();
         let mut handle = stdout.lock();
 
-        writeln!(handle, "{}", config.z).unwrap();
+        writeln!(handle, "{}", config.z)?;
         if !self.no_define_aliases {
-            writeln!(handle, "{}", config.alias).unwrap();
+            writeln!(handle, "{}", config.alias)?;
         }
 
         match self.hook {
             Hook::none => (),
-            Hook::prompt => writeln!(handle, "{}", config.hook.prompt).unwrap(),
+            Hook::prompt => writeln!(handle, "{}", config.hook.prompt)?,
             Hook::pwd => {
                 if let Some(pwd_hook) = config.hook.pwd {
-                    writeln!(handle, "{}", pwd_hook).unwrap();
+                    writeln!(handle, "{}", pwd_hook)?;
+                } else {
+                    bail!("PWD hooks are currently unsupported on this shell.");
                 }
             }
-        };
+        }
+
+        Ok(())
     }
 }
 
--- a/src/subcommand/init.rs
+++ b/src/subcommand/init.rs
@@ -29,6 +29,7 @@ impl Init {
         let config = match self.shell {
             Shell::bash => BASH_CONFIG,
             Shell::fish => FISH_CONFIG,
+            Shell::posix => POSIX_CONFIG,
             Shell::zsh => ZSH_CONFIG,
         };
 
@@ -43,7 +44,11 @@ impl Init {
         match self.hook {
             Hook::none => (),
             Hook::prompt => writeln!(handle, "{}", config.hook.prompt).unwrap(),
-            Hook::pwd => writeln!(handle, "{}", config.hook.pwd).unwrap(),
+            Hook::pwd => {
+                if let Some(pwd_hook) = config.hook.pwd {
+                    writeln!(handle, "{}", pwd_hook).unwrap();
+                }
+            }
         };
     }
 }
@@ -54,6 +59,7 @@ arg_enum! {
     enum Shell {
         bash,
         fish,
+        posix,
         zsh,
     }
 }
@@ -73,7 +79,7 @@ const BASH_CONFIG: ShellConfig = ShellConfig {
     alias: BASH_ALIAS,
     hook: HookConfig {
         prompt: BASH_HOOK_PROMPT,
-        pwd: BASH_HOOK_PWD,
+        pwd: Some(BASH_HOOK_PWD),
     },
 };
 
@@ -82,7 +88,16 @@ const FISH_CONFIG: ShellConfig = ShellConfig {
     alias: FISH_ALIAS,
     hook: HookConfig {
         prompt: FISH_HOOK_PROMPT,
-        pwd: FISH_HOOK_PWD,
+        pwd: Some(FISH_HOOK_PWD),
+    },
+};
+
+const POSIX_CONFIG: ShellConfig = ShellConfig {
+    z: POSIX_Z,
+    alias: POSIX_ALIAS,
+    hook: HookConfig {
+        prompt: POSIX_HOOK_PROMPT,
+        pwd: None,
     },
 };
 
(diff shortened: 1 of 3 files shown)


emersion:
>Is it possible to hook into when the prompt is redrawn, or when the pwd changes, in POSIX shells? 

`PS1` is evaluated when the prompt is redrawn. Maybe this can work, however the user might overwrite it.

cole-h:
I've implemented the prompt-redraw hook by using `PS1`. Only thing I don't like about it is, as @emersion noted above, users can potentially overwrite it. An alternative would be to just tell the user to add it themselves (and thus we won't have to worry about setting their prompt for them).

I was unable to make `_ZO_PWD` available to the shell itself (for use in the `PWD` hook), with or without `export`ing it. I'd appreciate any guidance on how I should go about this (or if it's even possible -- neither `mrsh` nor `bash --posix` appeared to work with or without the `export`).

---

Slighly related: @emersion `mrsh` might have a bug: running `zi` (`alias zi='z -i'`) did not display the `fzf` interactive picker thing, while in `bash --posix`, it did. A manual run of `zoxide query -i` worked in both. Is this a bug I should file on your tracker, or is this `bash --posix` actually doing something wrong?

cole-h:
Alright, I was able to implement a `PWD` hook. I don't really like that it requires creating a file and sourcing it, but that was the only way I could think of, other than just not having a `PWD` hook for POSIX shells. If this can be improved to not need a temporary file, I'd love to hear about it.

---

The only problems that I want to be addressed now are:

1) Are we fine with overwriting the PS1 if it's unset, or do we just substitute `PS1` in place and if it's unset, it's unset?

2) If `z -i` not working in `mrsh` is a bug or not (it works in `bash --posix`).

src/subcommand/init.rs
```diff
@@ -119,12 +130,12 @@ z() {
     if [ "$#" -eq 0 ]; then
         _z_cd ~ || return "$?"
     elif [ "$#" -eq 1 ] && [ "$1" = '-' ]; then
-        _z_cd ~- || return "$?"
```
ajeetdsouza: Removing this will cause the new path to be echoed in some shells. This essentially means that for those who have enabled `_Z_ECHO`, the new path will be echoed twice.

cole-h: This is either a Bashism, or a bug in `mrsh` (I should get around to filing these sometime soon). I can't find any documentation about `cd ~-` other than in the context of `bash`, so (for now and until this is made clear) wouldn't `cd "$@" >/dev/null` above do the same thing? Errors would still print, so I personally don't see a problem with it.

emersion: Yeah, `cd ~-` isn't POSIX.

emersion: `~-` is substituted by bash with `$OLDPWD` it seems.

ajeetdsouza: @emersion ideally, we should be able to replace this with `cd ${OLDPWD}` with no problems, correct?

emersion: Yeah, I think these are equivalent.

emersion:
>Is this a bug I should file on your tracker

Please file this bug, we'll investigate from there.

src/subcommand/init.rs
cole-h: Except for the case where `$OLDPWD` is null because it hasn't been set yet -- `cd -` in `bash` and `mrsh` report an error (`cd: OLDPWD is not yet` and similar), while `dash` appears to just `cd .` (thus setting `$OLDPWD`).

emersion: Seems like this isn't correct with `~-` either:

```
> bash
$ echo ~-
~-
$ cd ~-
```

cole-h: I've opted to branch based upon the value `$OLDPWD` (or lack thereof), allowing us to report an error in the case that it is unset (like `mrsh` and `bash --posix` do).

cole-h:
Before I mark this "ready for review," I'd like to know if we're OK with "hijacking" an unset prompt. If it's set, we do nothing (even if it's empty/null). If it's unset, we set the `PS1` to display as `$`. Are there any shortcomings I'm unaware of in this scenario? (Barring the argument "the user could change their `PS1`" -- which is valid, but there is nothing that we can do about it AFAICT.)

ajeetdsouza:
Why not let `PS1` remain an empty prompt if initially unset?

cole-h:
`mrsh` displays the prompt as `$` (or `#` if root) when `PS1` is unset, but if one were to `eval $(zoxide init posix)`, the prompt would be erased. Although I guess we could put that on them for not setting the PS1 initially. I'll file an issue for that, I think, and just substitute PS1, set, unset, or null.

cole-h:
This should generally be good to go. There might be a few nitpicks on the README, but that's easily dealt with.

I would like to wait for a decision in regards to https://github.com/emersion/mrsh/issues/149 before merging this, though. The wording of the standard is vague, so if that issue is closed without changes, it might be worth it to bring back the `${PS1-\\\$ }` substitution. It is standard to have the default `PS1` display as "$ ", anyways (see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_03).

ajeetdsouza:
I will try to look into this once I have more time - I feel like there must be a way to do this that doesn't involve using yet another file.

src/subcommand/init.rs
```diff
@@ -244,6 +275,26 @@ function _zoxide_hook --on-variable PWD
 end
 "#;
 
+// FIXME(cole-h): This is hacky, but the only way I was able to get it to work.
+const POSIX_HOOK_PWD: &str = r#"
+_zoxide_hook() {
+    touch /tmp/.zo_pwd
```
ajeetdsouza: A filename like this has problems:

- Multiple users will overwrite each others tmp files
- Multiple shell instances will overwrite each others tmp files
- `/tmp` is not always the preferred tmp directory

`mktemp` will create a tmp file for us in a compliant way, that won't clash with any other tmp file. We can this to generate a new file for each shell instance. However, `mktemp` is not part of the POSIX standard. While it is present on most modern operating systems, I still feel that using it defeats the purpose of POSIX compliance.

We could write our own logic for generating a tmp filename. However, we will end up cluttering the users tmp directory in no time, since every new shell will create a new tmp file - so we will also need to clean up upon shell exit by trapping `EXIT`, `INT`, `HUP`, and `TERM`.

It seems we really have opened a can of worms here - it may make sense to leave out the PWD hook as something we won't support in POSIX shells.

cole-h: (Purposefully not marking this as "resolved" so anybody who comes by here looking for answers as to why we didn't add a PWD hook can see the reasoning without needing to expand all the previously-resolved reviews.)

cole-h:
Anybody who is interested in implementing their own PWD hook, here was my naive approach. Feel free to modify it as necessary. Note that this has a variety of drawbacks and will not be officially supported. See the above comment by @ajeetdsouza for more information: https://github.com/ajeetdsouza/zoxide/pull/43#pullrequestreview-383447094

```sh
_zoxide_hook() {
    touch /tmp/.zo_pwd
    _ZO_PWD="$(cat /tmp/.zo_pwd)"

    if [ -z "${_ZO_PWD}" ]; then
        echo "${PWD}" > /tmp/.zo_pwd
    elif [ "${_ZO_PWD}" != "${PWD}" ]; then
        echo "${PWD}" > /tmp/.zo_pwd
        zoxide add
    fi
}

case "$PS1" in
    *\$\(_zoxide_hook\)*) ;;
    *) PS1="\$(_zoxide_hook)${PS1}" ;;
esac
```

src/subcommand/init.rs
```diff
@@ -43,7 +44,11 @@ impl Init {
         match self.hook {
             Hook::none => (),
             Hook::prompt => writeln!(handle, "{}", config.hook.prompt).unwrap(),
-            Hook::pwd => writeln!(handle, "{}", config.hook.pwd).unwrap(),
+            Hook::pwd => {
+                if let Some(pwd_hook) = config.hook.pwd {
```
ajeetdsouza: If the user has set `--hook pwd` on an unsupported shell, we should explicitly `bail!` and let them know, eg. `PWD hooks are currently unsupported on this shell`.

cole-h: That would require this function returning a `Result`. Are you fine with that?

ajeetdsouza: Absolutely!

src/subcommand/init.rs
```diff
@@ -36,20 +38,24 @@ impl Init {
         let stdout = io::stdout();
         let mut handle = stdout.lock();
 
-        writeln!(handle, "{}", config.z).unwrap();
+        writeln!(handle, "{}", config.z)?;
```
ajeetdsouza: I think this should be left as `unwrap`. All the errors we return have a user-friendly `context` attached to them, but it doesn't seem useful to do that here.

Rather, if writing to `stdout` fails, we can (reasonably) assume the user's computer is on fire and panic.

src/subcommand/init.rs
```diff
@@ -36,20 +38,24 @@ impl Init {
...
-            writeln!(handle, "{}", config.alias).unwrap();
+            writeln!(handle, "{}", config.alias)?;
         }
 
         match self.hook {
             Hook::none => (),
-            Hook::prompt => writeln!(handle, "{}", config.hook.prompt).unwrap(),
+            Hook::prompt => writeln!(handle, "{}", config.hook.prompt)?,
             Hook::pwd => {
                 if let Some(pwd_hook) = config.hook.pwd {
```
ajeetdsouza: I'd prefer a `match` clause over here - they tend to be more readable when you want to support both cases.

src/subcommand/init.rs
```diff
@@ -212,6 +241,17 @@ function _zoxide_hook --on-event fish_prompt
 end
 "#;
 
+const POSIX_HOOK_PROMPT: &str = r#"
+_zoxide_hook() {
+    zoxide add
```
ajeetdsouza: Given that anything this hook writes to `stdout` becomes part of the user's prompt (including debug statements, if someone is ever working on `zoxide` with a POSIX shell), I think it would be wise to redirect `stdout` elsewhere (either to `/dev/null` or `stderr`).

cole-h: I'll proceed under the assumption that we don't want to redirect `stderr` to `/dev/null`.

src/subcommand/init.rs
ajeetdsouza: I've added a PWD hook for POSIX shells in https://github.com/ajeetdsouza/zoxide/commit/7532fe98d88532022a95d387c4d21928b7026336, so I'm marking this as resolved.

jeff-hykin:
> Anybody who is interested in making their own PWD hook, here was my naive approach. Feel free to modify it as necessary. Note that this has a variety of drawbacks and will not be officially supported. See the above comment by @ajeetdsouza for more information: [#43 (review)](https://github.com/ajeetdsouza/zoxide/pull/43#pullrequestreview-383447094)

Thank you for this! After Stack overflow failed me, I knew Zoxide proved something like this was possible. I skimmed through the source, couldn't find much. Started digging through issues, and eventually I stumbled on this one. 👍

Issue #45, opened 2020-03-25
zoxide v0.3 - breaking changes


ajeetdsouza:
I'm planning to introduce some breaking changes to the way zoxide works in the next release. I will put my ideas into the comments below so they can be addressed separately. The idea is that we should do this as little as possible, so we should club the changes to ensure that users are least inconvenienced.

ajeetdsouza:
Instead of using a `String` to represent file paths, we should use the more correct `PathBuf` instead. To do string comparison, we use `to_string_lossy`, with which we can do case insensitive string comparison using [unicase](https://crates.io/crates/unicase), and (hopefully) fuzzy matching too.

ajeetdsouza:
All `zoxide` files should be stored under `"$XDG_CACHE_HOME/zoxide"` using [dirs](https://crates.io/crates/dirs).

ajeetdsouza:
We should introduce a persistent lock file too, which we can call `flock` on to ensure that `zoxide` is thread-safe, using [fs2](https://crates.io/crates/fs2).

ajeetdsouza:
We may want to consider moving away from [bincode](https://crates.io/crates/bincode) for serialization, because although it is the fastest Serde-based serializer/deserializer, it often aborts with a rather unintuitive message (`memory allocation of x bytes failed`) when there is a database corruption.

[This](https://github.com/erickt/rust-serialization-benchmarks) is a useful resource for rust serializers.

ajeetdsouza:
We may want to switch the `Rank` type from `f64` to `i32`. The advantage of using `f64` is that we get more "accurate" ranks, but I do not feel this would impact the algorithm significantly. However, we currently have to convert them to `i64` before sorting by rank due to the fact that floating-point values in Rust do not implement `Ord`.

A nice alternative to this would be to find a nicer way to floating point comparisons.

ajeetdsouza:
Currently, the internal representation of the DB is a `Vec<Dir>`. An alternative to this would be using `HashMap<PathBuf, DirStat>`, which would allow much simpler (and hopefully faster) database additions. Queries would be unaffected.

cole-h:
I can't wait for the switch to `PathBuf`s! I also can't wait for XDG Base Dir spec conformance!

[...]

Issue #44, opened 2020-03-21
Exclude certain directories from the database


cole-h:
Something that `z.lua`, `z.sh`, and `z` (fish) all have is a way to exclude/blacklist specific directories from the database. This would be useful to ensure that my home directory doesn't have an enormous rank (due to it being where my shell starts up initially), as well as hiding potentially sensitive/hidden/otherwise secret directories that should not be recorded.

Every time I open up `zi` and see my home directory, I am reminded of this, so I wanted to submit a formal (tracking) issue for it.

cole-h:
For anybody watching this: I'll probably work on this over the weekend.

cole-h:
#48 has been merged! However, only into the `v0.3` branch, which still has some issues that need to be addressed before it can be merged into `master`. I'll (try to remember to) make another comment here and close this issue once v0.3 releases.

Issue #39, opened 2020-03-17
use XDG compatible locations for the database/configs?


qmx:
It'd be nice to use standard config folders for this tool instead of just dropping configs and the database at the home folder for a user.

if you're interested about having this implemented, crates [like this one](https://crates.io/crates/dirs) can be used to accomplish this easily.

Happy to work on a PR if there's interest.

qmx:
also this crate I mentioned solves #7 in a cross-platform way

ajeetdsouza:
This has been discussed in https://github.com/ajeetdsouza/zoxide/issues/16, our conclusion was not to change anything for now as it breaks current setups.

If a major change in the future breaks compatibility unavoidably, I would be open to doing this.

ajeetdsouza:
This has been added in https://github.com/ajeetdsouza/zoxide/pull/46/commits/0617571807c461df094ad5e8dc95a154a587db65, and will be released with `zoxide v0.3`!

PR #46, opened 2020-03-25
zoxide v0.3

ajeetdsouza:
Closes https://github.com/ajeetdsouza/zoxide/issues/45, closes https://github.com/ajeetdsouza/zoxide/issues/44, closes https://github.com/ajeetdsouza/zoxide/issues/42, closes https://github.com/ajeetdsouza/zoxide/issues/39, closes https://github.com/ajeetdsouza/zoxide/issues/16, closes https://github.com/ajeetdsouza/zoxide/issues/7.


cole-h:
For the entire `0.3.x` version, I think it would be nice to have a notice about the changed db schema if we detect the user trying to load a db with the `0.2.x` schema. We can drop it after the next version bump. I'm sure it would already error, but it would be nice for it to be user friendly and describe exactly what needs to be done to become compatible with `0.3.x`.

Also, I think it would be better to have the upgrade functionality built into `zoxide` rather than use an external script for it. `zoxide migrate --upgrade-to-new-schema` (verbosity intentional), which takes no arguments and upgrades the db in-place, or something similar. Plenty of "are you sure?" prompts. The whole kit and caboodle. The in-place conversion would probably be done by rounding the scores up to the nearest integer.

src/util.rs
```diff
@@ -41,7 +42,12 @@ pub fn fzf_helper(now: Epoch, mut dirs: Vec<Dir>) -> Result<Option<String>> {
         dir.rank = dir.get_frecency(now);
     }
 
-    dirs.sort_by_key(|dir| std::cmp::Reverse(dir.rank as i64));
+    dirs.sort_unstable_by(|dir1, dir2| {
+        dir1.rank
+            .partial_cmp(&dir2.rank)
+            .unwrap_or(Ordering::Equal)
```
ajeetdsouza: I'm not 100% satisfied with the `.unwrap_or(Ordering::Equal)`, but we aren't expecting any incomparable values, and it's too trivial an issue to stop execution and throw an error. We do want to handle incomparable `f64` values elsewhere, though, presumably with `is_finite`, to ensure that they don't stick around in the database forever.

ajeetdsouza:
@cole-h, due to the change in DB location, `zoxide` would not try to load a DB from `~/.zo`. Unless what you mean is that a user is manually moving the database to `~/.cache/zoxide`? I'm not sure how we would detect that, but I do feel like we should add a schema version tag to the database to avoid this problem in future!

cole-h:
Hah, I misunderstood. I thought you were talking about upgrading the schema from `f64` -> `i32`.

I see no problem with having a fallback path like I proposed in the original issue -- search for a database at `$XDG_CACHE_HOME/zo` first, `$HOME/.cache/zo` second, and `$HOME/.zo` last. If none of those exist, we would create it at `$XDG_CACHE_HOME/zo` (or `$HOME/.cache/zo` if that env var is undefined).

This way, there's no need to "migrate" from `0.2.x` to `0.3.x` (unless you do decide to change `f64` -> `i32`). This is what many programs do and I think is the best solution.

ajeetdsouza:
I did some research and found a way of performing string matching on `PathBuf`s without the overhead of using `to_string_lossy` - it turns out the [`bstr`](https://crates.io/crates/bstr) is meant for exactly this kind of use case. On UNIX systems, it provides guaranteed lossless zero-cost conversions to and from `Path`s. On Windows systems, we will have to ignore paths with invalid Unicode - which we currently do anyway.

However, it turns out that switching to `PathBuf` may not be such a great idea after all. The reason for this is that once we perform matching and find the correct path to `cd` into, we still need to communicate it back to the shell. So far, I've been using pipes to communicate, and printing a Unicode `String` is quite straightforward in Rust. However, to print a `PathBuf`, we would need to use the lossy `PathBuf::display()` function. A lossy function is useless, because any change in the path will cause `cd` to fail.

So, we will not be able to transition to `PathBuf`s unless:
- we find a way of communicating a `PathBuf` losslessly over a pipe, or
- we find an alternative way to losslessy communicate a `PathBuf` value to the parent shell

cole-h:
If `PathBuf` isn't a solution for storage/communication, what about `OsString`? We could use the `From<OsString> for PathBuf` impl when we want to operate on a `PathBuf`, but utilize `OsString` everywhere else.

We might also consider dropping `envy` for env var support and just use the built-in [`std::env::var_os`](https://doc.rust-lang.org/std/env/fn.var_os.html). Unless you have a specific reason for wanting to use this crate, it would cut at least one unnecessary dependency.

ajeetdsouza:
@cole-h `PathBuf` uses `OsString` under the hood, so both of them have the same issue - there is no way (or I haven't found one) to losslessly print them to `stdout` because they are not guaranteed to be valid Unicode. `BString` can't be used for the same reason.

cole-h:
It might be a good time to think about starting a semver changelog, to document the breaking changes (and other improvements), as well as requiring entries to be added to it for user-facing changes in future PRs.

ajeetdsouza:
I think that's a great idea, I will add one once v0.3 is ready.

ajeetdsouza:
@cole-h, I've removed the dependency on `envy` for environment variable parsing, since:

- It seems to be extremely slow
- It doesn't fit our use case very well

Hopefully, this should help you with the `exclude-dirs` feature!

src/db.rs
```diff
@@ -154,42 +192,52 @@ impl DB {
...
+            self.data.dirs.retain(|dir| dir.rank >= 1.0);
         }
 
         self.modified = true;
         Ok(())
     }
 
     pub fn query(&mut self, keywords: &[String], now: Epoch) -> Option<Dir> {
-        let (idx, dir) = self
+        let (idx, dir, frecency) = self
```
cole-h: `frecency` appears to be unused. If you plan on using it later, should be fine to just prefix it with an underscore. Otherwise, it should probably be removed.

ajeetdsouza: Didn't see that, thanks!

cole-h: @ajeetdsouza (I think "resolved" comments no longer ping, so hoping the mention does)

Would it be possible to do something like this, instead? I really don't like seeing that unused 3rd tuple element when all it's necessary for is getting the max.

```rust
        let (idx, dir) = self
            .data
            .dirs
            .iter()
            .enumerate()
            .filter(|(_, dir)| dir.is_match(&keywords))
            .max_by(|(_, dir1), (_, dir2)| {
                let frecency1 = dir1.get_frecency(now);
                let frecency2 = dir2.get_frecency(now);

                frecency1.partial_cmp(frecency2).unwrap_or(Ordering::Equal)
            })?;
```

ajeetdsouza: @cole-h, while this did cross my mind, it results in 2 frecency calculations per `Dir` to the best of my knowledge, which is why I didn't go for it.

cole-h: @ajeetdsouza (ping again to ensure this gets to you)

I just did an extremely simple test on the playground to see the generated code.

Here's the source for the current version: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dcbaf05896ecb011e2ef57fc7daf7e7b

And for my proposal: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7273d0c2d7d67796c0d9c0e77d4939c5

You can compare for yourself, but:

```shell
λ wc -l mir1 mir2
  1900 mir1 # proposed
  1951 mir2 # current
  3851 total
```

```shell
λ wc -l llvm1 llvm2
   9768 llvm1 # proposed
  10247 llvm2 # current
  20015 total
```

```shell
λ wc -l asm1 asm2
  4618 asm1 # proposed
  5181 asm2 # current
  9799 total
```

---

The above are from the debug builds, but I looked at the ASM release build and it was 424 (proposed) vs 464 (current).

ajeetdsouza: @cole-h, here's what I tried on my machine:
- create and add these directories into the database: `test1`, `test2`, `test3`, `test4`
- add a `println` or a `sleep` statement to `get_frecency`
- run a query: `zoxide query test`

You will notice that in the current case, there are 4 calls to `get_frecency`, one for each directory in the database. However, the proposed method has 6 calls (`test1` & `test2`, `test2` & `test3`, and so on). Other than the first and last matches, each directory makes 2 calls to `get_frecency`.

I too feel that the querying logic is quite a bit ugly, and I'd be open to figuring out ways of making it better, but making 2 calls to `get_frecency` per directory does seem wasteful.

cole-h: @ajeetdsouza 

Then we could do the following (ugly, but it's what I came up with a single minute of thought):

```rust
        let (idx, dir) = self
            .data
            .dirs
            .iter()
            .enumerate()
            .filter(|(_, dir)| dir.is_match(&keywords))
            .map(|(idx, dir)| (idx, dir, dir.get_frecency(now)))
            .max_by(|(_, _, frecency1), (_, _, frecency2)| {
                frecency1.partial_cmp(frecency2).unwrap_or(Ordering::Equal)
            })
            .map(|(idx, dir, _)| (idx, dir))?; // <---- new 
```

I'll do more thinking on this, but wanted to put it here because 2 minds are better than 1.

ajeetdsouza: @cole-h I don't see the benefit of the additional `map` call, since it seems to be more boilerplate than the original code. I think it's fine to use destructuring to ignore part of a tuple; it's a fairly common occurrence in Rust code. 

I don't think there's a way of simplifying the iterator itself, so I think we should leave this for the time being and approach it later on if a solution presents itself.

cole-h: Fair enough. Thanks for bikeshedding with me.

src/db.rs
```diff
@@ -259,8 +263,11 @@ impl DB {
     }
 
     fn get_path_tmp(&self) -> PathBuf {
+        let file_name = format!(".{}.zo", Uuid::new_v4());
```
cole-h: What problem is this solving? I personally don't like the idea of adding **8** new dependencies just to get a random filename.

src/db.rs
```diff
@@ -38,30 +39,27 @@ impl DB {
...
+            fs::rename(&path_tmp, &self.path).context("could not move temporary database file")?;
         }
 
         Ok(())
     }
 
     pub fn migrate<P: AsRef<Path>>(&mut self, path: P, merge: bool) -> Result<()> {
         if !self.dirs.is_empty() && !merge {
-            bail!(
+            bail!(indoc!(
```
cole-h: Similar to the above: I see no reason to use a crate that pulls in 4 transitive dependencies, when `"{}\n{}"` or the previous solution works just fine. Maybe this is me trying to be minimalist, but this is used only once: in this single line...

As you can probably gather, I'm not a fan of `left-pad`-like dependencies, even though Rust makes them so easy to use.

src/db.rs
ajeetdsouza: This was done in order to achieve thread-safety. Currently, there is no locking mechanism in place, which means multiple instances of `zoxide` could be writing to the same tmp file concurrently, which would corrupt the database.

My initial idea was to use `flock`s to ensure that only one transaction can take place at a time. However, that isn't a great idea, since a misbehaving `zoxide` instance would hang every shell instance on the system.

To prevent this, I'm now generating a new tmp filename for every transaction. This ensures that while some transactions may be lost, the database itself will never be corrupted.

The reason I've used UUIDs (and not an RNG) to generate the filename is because they are virtually guaranteed to be unique, which saves us the trouble of checking if there is already a tmp file by that name.

An alternative to this would be to use [`create_new`](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create_new), which will ensure that we don't accidentally write to a preexisting tmp file. Whether or not we remove UUIDs, I think `create_new` would be a good thing to have here.

src/db.rs
ajeetdsouza: The only overhead of `indoc` is at compile time - it does not have any impact on the output binary.

I included this primarily because I was hoping to use it for https://github.com/ajeetdsouza/zoxide/issues/32. The issue there is that I will have to replace the large static strings by formatted strings, generated inside functions. I would really have liked indented strings there, because not having them really screws up code collapsing on my editor, which is frustrating (especially for large functions/strings).

The venture proved pointless, however - `indoc` only supports formatted strings on nightly Rust. For now, since there is no runtime impact, I'm thinking of leaving it in all the same (it might prove useful elsewhere), but I would gladly hear your take on the matter.

cole-h: I don't disagree that there is no runtime impact. However, there is a compile-time impact -- 4 more crates need to be compiled and linked. Sure, this isn't much of a problem now, but it might become one down the line. As implied in my original comment, dependencies that can be implemented solely using the standard library irk me.

If it does prove useful elsewhere, then that would be a perfect opportunity to add it! But keeping it around "just in case" seems like a waste to me.

ajeetdsouza: That's true, I will remove the dependency.

cole-h:
Just a small note for the future: I think it would be better to close issues only after they have been resolved on `master` (not just when the PR has been merged into another branch). Just add it to the PR message with a `Fixes #<num>.` and GitHub should close it for you (like it did with the POSIX shell support PR).

That way people know their issue is solved and they don't have to now wait for the next release (which, admittedly, doesn't seem too far off).

ajeetdsouza:
That's a good idea, I will do this the next time!

src/db.rs
cole-h: Still not happy about it, but it does solve an actual problem. I'll look into lighter alternatives, but that shouldn't block this getting merged.

`create_new` will error if the file already exists, so while it's a good addition, it's not a solution. Looks like `uuid` will be necessary to address this concern, at least for now.

ajeetdsouza:
@cole-h I've added a `CHANGELOG.md`. I think this should be ready to release.

ajeetdsouza:
@cole-h I've added a `CHANGELOG.md`. I think this should be ready to release.

CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.3.0] - 2020-03-30
+
+### Added
+
+- Added `CHANGELOG.md`
```
cole-h:
```suggestion
```

"The purpose of a changelog entry is to document the noteworthy difference, often across multiple commits, to communicate them clearly to end users. "

I don't consider a changelog "noteworthy", because it doesn't impact our users :P

CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
...
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.3.0] - 2020-03-30
+
+### Added
+
```
cole-h:
```suggestion
## [Unreleased]

## [0.3.0] - 2020-03-30
### Added
```

CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
...
+- Added `CHANGELOG.md`
+- Added `$_ZO_EXCLUDE_DIRS` to prevent certain directories from being added to the database.
+- Added support for generic POSIX shells.
+
+### Changed
+
+- Switched default database location to user's data directory.
+- Upgraded database schema to include a version number, and added automatic migration from `v0.2.x` databases (will be removed in `v0.4.0`).
```
cole-h:
```suggestion
- Upgraded database schema to include a version number, and added automatic migration from `v0.2.x` databases.
```

As above, changelog entries should describe user-facing changes. The fact that it will be removed later is not a user-facing change -- it will go under the `### Removed` header whenever it happens.

CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Added
+
+- Added `CHANGELOG.md`
+- Added `$_ZO_EXCLUDE_DIRS` to prevent certain directories from being added to the database.
+- Added support for generic POSIX shells.
+
+### Changed
+
+- Switched default database location to user's data directory.
```
cole-h: I think it looks nicer (when viewing it as raw text) to have the header be directly above what it is describing. If you're fine with it, I'd like to ask that all headers follow this same kind of decoration: Header, directly followed by entries, with a newline separating the next header.

```suggestion
### Changed
- Switched default database location to user's data directory.
```

CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
...
+- Upgraded database schema to include a version number, and added automatic migration from `v0.2.x` databases (will be removed in `v0.4.0`).
+- Changed `migrate` subcommand to `import`.
+
+### Fixed
+
+- Used unique temporary database file names to achieve thread safety.
+- Fixed incomprehensive "could not allocate" message on database corruption.
```
cole-h: I'm not gonna go through every single entry, but the entries should be read like: `<Header>: <entry>`, e.g. `Fixed: Incomprehensive "could not allocate" message on database corruption.`

```suggestion
- Achieve thread safety by using a unique, temporary database file.
- Incomprehensive "could not allocate" message on database corruption.
```

CHANGELOG.md
```diff
@@ -0,0 +1,91 @@
...
+
+## [0.2.1] - 2020-03-16
+
+### Added
+
+- Added `$_ZO_ECHO` environment variable to echo match before `cd`ing.
+- Added a minimal `ranger` plugin.
+- Added a PWD hook to only update the database when the current directory is changed.
+- Added support for the `bash` shell.
+- Added the `migrate` subcommand to allow users to migrate from `z`.
```
cole-h: See the above comment.

```suggestion
- `$_ZO_ECHO` environment variable to echo match before `cd`ing.
- Minimal `ranger` plugin.
- PWD hook to only update the database when the current directory is changed.
- Support for the `bash` shell.
- `migrate` subcommand to allow users to migrate from `z`.
```

src/config.rs
```diff
@@ -0,0 +1,54 @@
...
+    let data_dir = match env::var_os("_ZO_DATA_DIR") {
+        Some(data_osstr) => PathBuf::from(data_osstr),
+        None => {
+            if let Some(mut data_dir) = dirs::data_local_dir() {
+                data_dir.push("zoxide");
+                data_dir
+            } else {
+                bail!("could not find database directory, please set _ZO_DATA_DIR manually");
+            }
+        }
```
cole-h: You asked me to `match` on something similar to this in my rework of the fallback. Maybe we should do the same here (feel free to rebase it into the rework commit).

```suggestion
        None => match dirs::data_local_dir() {
            Some(mut data_dir) => {
                data_dir.push("zoxide");
                data_dir
            None => bail!("could not find database directory, please set _ZO_DATA_DIR manually"),
        }
```

src/db.rs
```diff
@@ -1,67 +1,111 @@
...
             Err(err) => match err.kind() {
-                io::ErrorKind::NotFound => Vec::<Dir>::new(),
-                _ => return Err(err).with_context(|| anyhow!("could not open database file")),
+                io::ErrorKind::NotFound => DBData::default(),
+                _ => return Err(err).context("could not open database file"),
             },
         };
 
+        if data.version != config::DB_VERSION {
+            bail!("this database version ({}) is unsupported", data.version);
```
cole-h: Get right to the point.

```suggestion
            bail!("database version '{}' is unsupported", data.version);
```

src/db.rs
```diff
@@ -1,67 +1,111 @@
...
 
+pub use i32 as DBVersion;
+
 pub struct DB {
-    path: PathBuf,
-    dirs: Vec<Dir>,
+    data: DBData,
     modified: bool,
+    path: PathBuf,
+    path_old: Option<PathBuf>,
```
cole-h: I don't think we're planning on moving DB locations ever again after this, so might as well make a note to remove this when it's no longer needed (next breaking release).

```suggestion
    // FIXME: remove after next breaking version
    path_old: Option<PathBuf>,
```

src/db.rs
```diff
@@ -1,67 +1,111 @@
...
-            path,
-            dirs,
+            data,
             modified: false,
+            path: path.as_ref().to_path_buf(),
+            path_old: None,
+        })
+    }
+
+    pub fn open_and_migrate<P1, P2>(path_old: P1, path: P2) -> Result<DB>
```
cole-h: Making notes so nobody forgets, just in case.

```suggestion
    // FIXME: remove after next breaking version
    pub fn open_and_migrate<P1, P2>(path_old: P1, path: P2) -> Result<DB>
```

src/db.rs
```diff
@@ -1,67 +1,111 @@
...
 
     pub fn save(&mut self) -> Result<()> {
         if self.modified {
             let path_tmp = self.get_path_tmp();
 
-            let file_tmp = File::create(&path_tmp)
-                .with_context(|| anyhow!("could not open temporary database file"))?;
+            let file_tmp = OpenOptions::new()
+                .write(true)
+                // ensure that we are not overwriting an existing file_tmp
```
cole-h:
```suggestion
```

If you really want this comment, it should be above the assignment and something along the lines of (reflowing necessary):

```rust
// Calling `OpenOptions::new()` with `.create_new(true)` ensures that we are not overwriting an existing temporary file; it will return an error when the file already exists.
```

src/db.rs
```diff
@@ -219,38 +257,47 @@ If you wish to merge the two, specify the `--merge` flag."
...
+        path_tmp
     }
 }
 
 impl Drop for DB {
     fn drop(&mut self) {
         if let Err(e) = self.save() {
             eprintln!("{:#}", e);
+        } else if let Some(path_old) = &self.path_old {
+            if let Err(e) = fs::remove_file(path_old).context("could not remove old database") {
```
cole-h:
```suggestion
            // FIXME: remove this branch after next breaking release
            if let Err(e) = fs::remove_file(path_old).context("could not remove old database") {
```

src/dir.rs
```diff
@@ -1,36 +1,50 @@
-use crate::types::{Epoch, Rank};
 use serde::{Deserialize, Serialize};
-use std::path::Path;
+use std::path::PathBuf;
+
+pub use f64 as Rank;
+pub use i64 as Epoch; // use a signed integer so subtraction can be performed on it
```
cole-h: I think this comment is unnecessary. Just a suggestion; feel free to disregard if you think otherwise.

```suggestion
pub use i64 as Epoch;
```

src/db.rs
ajeetdsouza: I'll mark this as resolved, since `uuid` can be replaced painlessly in future releases. Do create an issue if you find a better way of doing this.

src/util.rs
```diff
@@ -1,36 +1,60 @@
...
 
-pub fn get_db(env: &Env) -> Result<DB> {
-    let path = env
-        .data
-        .as_ref()
-        .ok_or_else(|| anyhow!("could not locate database file"))?;
-    DB::open(path)
+#[cfg(unix)]
+pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
+    use std::os::unix::ffi::OsStrExt;
```
cole-h: Nit.

```suggestion
    use std::os::unix::ffi::OsStrExt;
    
```

CHANGELOG.md
ajeetdsouza: I'm not sure about this - [`prettier`](https://prettier.io/), which is the Markdown formatter I use, adds in blanks before and after headers, and [`markdownlint`](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines) also seems to agree.

src/util.rs
cole-h: Do we need to handle incomparable `f64`s if we default to a `MAXAGE` of `1000.0`, clamp frecency to `9999.0`, and anything below `1.0` gets removed?

src/util.rs
```diff
@@ -89,3 +117,18 @@ pub fn fzf_helper(now: Epoch, mut dirs: Vec<Dir>) -> Result<Option<String>> {
...
+    assert!(min <= max);
+    let mut x = val;
+    if x < min {
+        x = min;
+    }
+    if x > max {
+        x = max;
+    }
+    x
+}
```
cole-h: Just to make it look nicer.

```suggestion
pub fn clamp(val: f64, min: f64, max: f64) -> f64 {
    assert!(min <= max);
    
    let mut x = val;
    
    if x < min {
        x = min;
    }
    
    if x > max {
        x = max;
    }
    
    x
}
```

CHANGELOG.md
cole-h: Fair enough. Even my "gold standard of changelogs" Alacritty agrees!

CHANGELOG.md
```diff
@@ -0,0 +1,93 @@
...
+### Added
+
+- Automatic migration from `v0.2.x` databases.
+- `$_ZO_EXCLUDE_DIRS` to prevent certain directories from being added to the database.
+- Support for generic POSIX shells.
+
+### Changed
+
+- Database location defaults to user's data directory.
+- Upgrade database schema to include a version number.
```
cole-h:
```suggestion
- Database location defaults to user's local data directory.
- Database schema now includes a version number.
```

CHANGELOG.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Fixed
+
+- Interactive queries causing other open shells to hang.
+
+## [0.2.0] - 2020-03-11
+
+### Added
+
+- `init` subcommand to remove dependency on shell package managers.
```
cole-h: Wording nit.

```suggestion
- `init` subcommand to remove dependency on shell plugin managers.
```

CHANGELOG.md
```diff
@@ -0,0 +1,93 @@
...
+## [0.1.1] - 2020-03-08
+
+### Added
+
+- Install script for Linux/macOS users.
+- Aging algorithm to remove stale entries.
+
+### Changed
+
+- Upgrade database schema to use `f64` values for rank rather than `i32`.
```
cole-h:
```suggestion
- Database schema now uses `f64` values for rank instead of `i32`.
```

CHANGELOG.md
```diff
@@ -0,0 +1,93 @@
...
+
+## [Unreleased]
+
+## [0.3.0] - 2020-03-30
+
+### Added
+
+- Automatic migration from `v0.2.x` databases.
+- `$_ZO_EXCLUDE_DIRS` to prevent certain directories from being added to the database.
+- Support for generic POSIX shells.
```
cole-h: Oh, and this, please.

```suggestion
- Support for POSIX-compliant shells.
```

src/dir.rs
ajeetdsouza: I left that in there because it can be tempting to change it to a `u64`, since that is what `Duration::as_secs` returns. However, this could potentially let errors go undetected, so it's better to be explicit here.

src/util.rs
```diff
@@ -1,36 +1,61 @@
...
-    DB::open(path)
+#[cfg(unix)]
+pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
+    use std::os::unix::ffi::OsStrExt;
+
+    Some(path.as_ref().as_os_str().as_bytes())
+}
+
+#[cfg(not(unix))]
+pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
```
cole-h: Why are these `Option`s? Just so you can use `to_str()?`?

ajeetdsouza: This operation can fail on Windows systems, returning an `Option` forces the caller to `unwrap` it explicitly, or handle the error.

PR #47, opened 2020-03-27
Rework database fallback for v0.3

cole-h:
Motivation behind this change is that the old version would only ever use the database at the data home, unless the data home folder (`$HOME/.local/share`) did not exist. This implementation checks for the existence of the old database at `$HOME/.zo`and migrates it to the new storage location; otherwise, it uses `$XDG_DATA_HOME/zoxide` (and creates it if it doesn't exist).

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,12 @@ zoxide init fish | source
 
 ### Environment variables
 
-- `$_ZO_DATA`: sets the location of the database (default: `~/.zo`)
+- `$_ZO_DATA_DIR`: directory where `zoxide` will store its data files (default:
+    platform-specific; see the [`dirs` documentation] for more information)
 - `$_ZO_ECHO`: `z` will print the matched directory before navigating to it
 - `$_ZO_EXCLUDE_DIRS`: list of directories separated by platform-specific
     characters (`:` on Linux and macOS, and `;` on Windows) to be excluded from
     the database
 - `$_ZO_MAXAGE`: sets the maximum total rank after which entries start getting deleted
+
+[`dirs` documentation]: https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,27 +8,26 @@ use std::fs;
 use std::path::PathBuf;
 
 pub const DB_MAX_SIZE: u64 = 8 * 1024 * 1024; // 8 MiB
-
 pub const DB_VERSION: DBVersion = 3;
 
-pub fn zo_data() -> Result<PathBuf> {
-    let path = match env::var_os("_ZO_DATA") {
+pub fn zo_data_dir() -> Result<PathBuf> {
+    let data_dir = match env::var_os("_ZO_DATA_DIR") {
         Some(data_osstr) => PathBuf::from(data_osstr),
         None => {
-            if let Some(mut cache_dir) = dirs::cache_dir() {
-                cache_dir.push("zoxide");
-                cache_dir
-            } else if let Some(mut home_dir) = dirs::home_dir() {
-                home_dir.push(".zoxide");
-                home_dir
+            if let Some(mut data_dir) = dirs::data_local_dir() {
+                data_dir.push("zoxide");
+                data_dir
             } else {
-                bail!("could not generate default directory, please set _ZO_DATA manually");
+                bail!("could not find database directory, please set _ZO_DATA_DIR manually");
             }
         }
     };
 
-    fs::create_dir_all(&path).context("could not create _ZO_DATA directory")?;
-    Ok(path)
+    // This will fail when `data_dir` points to a file or a broken symlink, but
+    // will no-op on a valid symlink (to a directory), or an actual directory.
+    fs::create_dir_all(&data_dir).context("could not create data directory")?;
+
+    Ok(data_dir)
 }
 
 pub fn zo_exclude_dirs() -> Vec<PathBuf> {
@@ -45,6 +44,7 @@ pub fn zo_maxage() -> Result<Rank> {
                 let maxage = maxage_str
                     .parse::<i64>()
                     .context("unable to parse _ZO_MAXAGE as integer")?;
+
                 Ok(maxage as Rank)
             }
             None => bail!("invalid Unicode in _ZO_MAXAGE"),
--- a/src/db.rs
+++ b/src/db.rs
@@ -48,7 +48,7 @@ impl DB {
         })
     }
 
-    pub fn open_old<P1, P2>(path_old: P1, path: P2) -> Result<DB>
+    pub fn open_and_migrate<P1, P2>(path_old: P1, path: P2) -> Result<DB>
     where
         P1: AsRef<Path>,
         P2: AsRef<Path>,
--- a/src/util.rs
+++ b/src/util.rs
@@ -22,9 +22,19 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
 }
 
 pub fn get_db() -> Result<DB> {
-    let mut path = config::zo_data()?;
-    path.push("db.zo");
-    DB::open(path)
+    let mut db_file = config::zo_data_dir()?;
+    db_file.push("db.zo");
+
+    // FIXME: fallback to old database location; remove in next breaking version
+    if let Some(mut old_db) = dirs::home_dir() {
+        old_db.push(".zo");
+
+        if old_db.is_file() && !db_file.is_file() {
+            return DB::open_and_migrate(old_db, db_file);
+        }
+    }
+
+    DB::open(db_file)
 }
 
 pub fn get_current_time() -> Result<Epoch> {
--- a/src/util.rs
+++ b/src/util.rs
@@ -22,19 +22,21 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
 }
 
 pub fn get_db() -> Result<DB> {
-    let mut db_file = config::zo_data_dir()?;
-    db_file.push("db.zo");
+    let mut db_path = config::zo_data_dir()?;
+    db_path.push("db.zo");
 
     // FIXME: fallback to old database location; remove in next breaking version
-    if let Some(mut old_db) = dirs::home_dir() {
-        old_db.push(".zo");
+    if !db_path.is_file() {
+        if let Some(mut old_db_path) = dirs::home_dir() {
+            old_db_path.push(".zo");
 
-        if old_db.is_file() && !db_file.is_file() {
-            return DB::open_and_migrate(old_db, db_file);
+            if old_db_path.is_file() {
+                return DB::open_and_migrate(old_db_path, db_path);
+            }
         }
     }
 
-    DB::open(db_file)
+    DB::open(db_path)
 }
 
 pub fn get_current_time() -> Result<Epoch> {


ajeetdsouza:
Looks like you beat me to the implementation!

So far, there have been no breaking changes to the schema, since the serializer seems to handle `PathBuf`s and `String`s the same way. However, I'm thinking of adding a `version` tag to the schema, to allow for seamless upgrades in the future.

Here's what I was planning to do:

- If `$_ZO_DATA` is set:
  - If `$_ZO_DATA/zo.db` does not exist and `$_ZO_DATA/.zo` exists:
    - Migrate it to the new schema and place it under `$_ZO_DATA/zo.db`, after which we delete `$_ZO_DATA/.zo`
- Otherwise:
  - If `<cache_dir>/zoxide/zo.db` does not exist and `$HOME/.zo` exists:
    - Migrate it to the new schema and place it under `<cache_dir>/zoxide/zo.db`, after which we delete `$_ZO_DATA/.zo`

Also, for next time, do try to separate PRs that implement different features!

cole-h:
I personally think it would make more sense to have `$_ZO_DATA` be the path to the database, rather than to the folder holding the database. This way, we follow the precedent that `z.sh`, `z` (fish), and `z.lua` have set (data env var pointing directly to the database) -- the principle of least surprise for all the users that are migrating from those alternatives. "Oh, I changed `_ZO_DATA` to be the same as `_Z_DATA`, why does it error saying that it's a file and not a folder?"

Maybe there's a reason you want to have `_ZO_DATA` point to a folder? If so, then maybe it would be better to rename to `_ZO_DATA_DIR` to more precisely describes its purpose.

I still want to put it in the data dir as opposed  to the cache dir ;)

---

I like the idea of tagging the DB with a version. I'm no DB wizard, so interested to see what you do with it.

---

I was so excited about getting this out, I didn't separate them. Will do for the next time ;^) Could still do it now -- I'd just move the EXCLUDE_DIRS stuff to a new PR in order to keep the on-topic discussion regarding database fallback here.

ajeetdsouza:
I agree with you - I think the data directory is a better choice than the cache directory.

However, I think it would be better to use a directory for storing the database, for the following reasons:
- If we should ever need more files (for example, a tmp file, or a lock file) along with the database, we don't want to pollute the data directory with the extra files. It would be much cleaner if it were all under one directory instead.
- Taking a look at my own `~/.local/share`, it seems to be the convention for most applications to group their data files under one directory.

I would not be opposed to renaming the environment variable to `_ZO_DATA_DIR`, though.

cole-h:
I'm not saying we shouldn't use a directory for storing the database and any other files, but that `_ZO_DATA` -- if a user wants to use it -- should point directly at the database. If it is set, then we use its parent directory to store the tempfile/lockfile (adjacent to the database). If it's not set, we use the (new) default of `~/.local/share/zoxide/{db,tempfile,lockfile}`.

Just clarifying what I was trying to communicate. Either way, I like the `_ZO_DATA_DIR` rename better; working on it.

src/util.rs
```diff
@@ -22,9 +22,18 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
...
 
 pub fn get_db() -> Result<DB> {
-    let mut path = config::zo_data()?;
-    path.push("db.zo");
-    DB::open(path)
+    // FIXME: fallback to old database location; remove in next breaking version
+    if let Some(home_dir) = dirs::home_dir() {
+        let db_file = home_dir.join(".zo");
+
+        if db_file.exists() && db_file.is_file() {
```
ajeetdsouza: The `exists` check can be removed here, since `is_file` does that check anyway.

ajeetdsouza:
More work is needed here regarding migration to the soon-to-be-changed schema. Let's keep this PR on hold till I am able to work on it.

src/util.rs
cole-h: Fixed locally. Once we settle on the schema changes, I'll push along with any other changes that'll need to be made.

ajeetdsouza:
@cole-h, I think you should be able to continue work on this. I've added an `open_old` method to the `DB` struct, which you can use to import a v0.2 database.

src/db.rs
```diff
@@ -47,7 +47,7 @@ impl DB {
         })
     }
 
-    pub fn open_old<P1, P2>(path_old: P1, path: P2) -> Result<DB>
+    pub fn open_and_migrate<P1, P2>(path_old: P1, path: P2) -> Result<DB>
```
cole-h: I renamed this to be more indicative of what it does -- it opens the old database and migrates it to the new, XDG-conforming location; it doesn't "just" open the old database.

README.md
```diff
@@ -101,7 +101,7 @@ zoxide init fish | source
 
 ### Environment variables
 
-- `$_ZO_DATA`: sets the location of the database (default: `~/.zo`)
+- `$_ZO_DATA_DIR`: directory where `zoxide` will store its data files (default: `$HOME/.local/share/zoxide`)
```
ajeetdsouza: `$HOME/.local/share/zoxide` isn't really accurate since it's only true for Linux systems. I don't have the answer to what you would replace it with without being needlessly verbose, though.

src/config.rs
```diff
@@ -8,27 +8,31 @@ use std::fs;
...
+            Ok(data_dir)
+        }
         None => {
-            if let Some(mut cache_dir) = dirs::cache_dir() {
-                cache_dir.push("zoxide");
-                cache_dir
-            } else if let Some(mut home_dir) = dirs::home_dir() {
-                home_dir.push(".zoxide");
-                home_dir
+            if let Some(xdg_data_dir) = dirs::data_dir() {
```
ajeetdsouza: Use `dirs::data_local_dir()` instead. On Windows, `dirs::data_dir()` defaults to `C:\Users\Alice\AppData\Roaming`, which is meant to be shared by the same user across different computers.

src/config.rs
```diff
@@ -8,27 +8,31 @@ use std::fs;
...
+                let data_dir = xdg_data_dir.join("zoxide");
```
ajeetdsouza: `xdg_data_dir` is a Linux-specific concept, and `join` unnecessarily creates a new `PathBuf`. Instead, why not use `mut data_dir` and perform the `push` operation on it?

src/config.rs
```diff
@@ -8,27 +8,31 @@ use std::fs;
...
 
-pub fn zo_data() -> Result<PathBuf> {
-    let path = match env::var_os("_ZO_DATA") {
-        Some(data_osstr) => PathBuf::from(data_osstr),
+pub fn zo_data_dir() -> Result<PathBuf> {
+    match env::var_os("_ZO_DATA_DIR") {
+        Some(data_osstr) => {
+            let data_dir = PathBuf::from(data_osstr);
+
+            if data_dir.is_file() {
```
ajeetdsouza: There's a problem here - even if `data_dir` is not a file, it still need not be a directory (eg. a symlink to a file).

Instead, I'd suggest returning a `PathBuf` from both branches of the `match` expression, and performing `fs::create_dir_all` on it (which should error out if `data_dir` is anything but a directory).

src/db.rs
ajeetdsouza: Fair enough!

README.md
cole-h: We could just cut out the "default" verbiage, if you're fine with it. Otherwise, I'll brainstorm something better.

src/config.rs
cole-h: I always forget you want to target Windows as well; sorry. I'll revert back to how it was before (`mut data_dir`).

src/config.rs
cole-h: I'm not entirely grokking the problem (through not fault of your own -- a little tired at the moment). I'll revisit this in the morning.

README.md
ajeetdsouza: That seems fine to me.

src/config.rs
cole-h: OK, I understand now. I added a comment above the `fs::create_dir_all` call to document its use.

README.md
cole-h: I actually changed the "default:" wording to point to the `dirs` docs. Is this a satisfactory solution, or are you still leaning more towards dropping it altogether?

src/util.rs
```diff
@@ -22,9 +22,19 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
 }
 
 pub fn get_db() -> Result<DB> {
-    let mut path = config::zo_data()?;
-    path.push("db.zo");
-    DB::open(path)
+    let data_dir = config::zo_data_dir()?;
+    let db_file = data_dir.join("db.zo");
```
ajeetdsouza: Avoidable `join` - try replacing it with a `push`.

src/util.rs
```diff
@@ -22,9 +22,19 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
...
+
+    // FIXME: fallback to old database location; remove in next breaking version
+    if let Some(home_dir) = dirs::home_dir() {
+        let old_db = home_dir.join(".zo");
```
ajeetdsouza: Avoidable `join` - try replacing it with a `push`.

README.md
```diff
@@ -101,9 +101,12 @@ zoxide init fish | source
 
 ### Environment variables
 
-- `$_ZO_DATA`: sets the location of the database (default: `~/.zo`)
+- `$_ZO_DATA_DIR`: directory where `zoxide` will store its data files (default:
+    platform-specific; see the [`dirs` documentation] for more information)
```
ajeetdsouza: I think it would be better to use this link (`https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html`) instead, so that it stays up-to-date.

ajeetdsouza: For brevity, I think it would be better to change this line to "see here for more" and include an inline link.

src/util.rs
cole-h: Forgot to go through the other part and replace these joins, whoops.

README.md
cole-h: I'll change it to point to latest.

By "see here for more" and inline link, do you mean: `see [here](insert url here) for more`? Or do you mean: `see here for more: <insert url here>`?

The current situation is already an inline link; rendered: https://github.com/ajeetdsouza/zoxide/blob/52ae1b03bcdcb148904936bc6d6f5c0ebd334372/README.md#environment-variables.

ajeetdsouza: I think `see [here](insert url here) for more` would be better. By inline link, I meant that if the text is changed to `here`, it wouldn't make much sense to separate the link out. Should have been more clear, my bad!

cole-h: All good. But I don't agree that `see [here](url) for more` is better than `see the [dirs doumentation] for more`, even in the interest of brevity. We don't have nearly enough doc comments to excuse short documentation in the readme, IMHO.

`see [here]` just sends you blindly to another page (ignore the fact that hovering shows URLs is most modern browsers); "where is 'here'? Why won't they tell me what this link is? Am I going  to be sent to a phishing page? Do I trust this project?" Hyperbolic, but hopefully it gets my point across.

`see the [dirs documentation]`, on the other hand, tells you exactly where you're going: to read the `dirs` documentation.

ajeetdsouza: That's a fair argument!

src/util.rs
```diff
@@ -22,9 +22,19 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
...
 
 pub fn get_db() -> Result<DB> {
-    let mut path = config::zo_data()?;
-    path.push("db.zo");
-    DB::open(path)
+    let mut db_file = config::zo_data_dir()?;
+    db_file.push("db.zo");
+
+    // FIXME: fallback to old database location; remove in next breaking version
+    if let Some(mut old_db) = dirs::home_dir() {
```
ajeetdsouza: This just occurred to me - if the user already has a v0.3 database, it will get overwritten. I think we should only perform the migration if `db_file` is not a file.

cole-h: Good catch, thanks!

cole-h: Just as a note: you can't combine `if let` with a boolean, so that's why I had to check if `old_db` is a file and `db_file` is not a file.

src/util.rs
```diff
@@ -22,9 +22,19 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
...
+        old_db.push(".zo");
+
+        if old_db.is_file() && !db_file.is_file() {
```
ajeetdsouza: While this does increase the level of nesting, we only need to compute `old_db` in case `db_file` is not a file. How about something like:

```rust
if !db_file.is_file() {
    if let Some(mut old_db) =  = dirs::home_dir() {
        todo!()
    }
}
```

src/util.rs
```diff
@@ -22,9 +22,19 @@ pub fn path_to_bytes<P: AsRef<Path>>(path: &P) -> Option<&[u8]> {
 }
 
 pub fn get_db() -> Result<DB> {
-    let mut path = config::zo_data()?;
-    path.push("db.zo");
-    DB::open(path)
+    let mut db_file = config::zo_data_dir()?;
```
ajeetdsouza: The variable name here is misleading, since this is not a file. How about something like `db_path`?

src/util.rs
ajeetdsouza: For consistency (with `db_path`), I think a name `old_db_path` would be more suitable here.

ajeetdsouza:
Also, a small note - since I'm squash-merging PRs, there isn't any need to force-push your commits. In fact, regular commits are more helpful, since I can see exactly what's been changed since the last review!

src/util.rs
cole-h: Lots of nesting, but it's unavoidable.

src/util.rs
cole-h: Not a file *yet*... Point taken; done.

cole-h:
I've been trying to be better about that. Some projects (like this and Alacritty) I can count on to squash-merge; others merge all my in-between commits. Now that I know I can count on you to do the right thing, I'll (try to) do the right thing, too.

ajeetdsouza:
This is perfect, thank you!

PR #48, opened 2020-03-28
Implement _ZO_EXCLUDE_DIRS for v0.3

cole-h:
_ZO_EXCLUDE_DIRS is a list of paths (separated by colons, `:`, on
Unix-based systems, and semicolons, `;`, on Windows) that should be
excluded from the database. Example:

    _ZO_EXCLUDE_DIRS="$HOME:$HOME/something/super/secret:$HOME/caused/by/background/cds"

---

Closes https://github.com/ajeetdsouza/zoxide/issues/44.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,7 @@
 use crate::types::Rank;
 
 use anyhow::{bail, Context, Result};
+
 use std::env;
 use std::fs;
 use std::path::PathBuf;
@@ -23,6 +24,13 @@ pub fn zo_data() -> Result<PathBuf> {
 
     fs::create_dir_all(&path).context("could not create _ZO_DATA directory")?;
     Ok(path)
+}
+
+pub fn zo_exclude_dirs() -> Vec<PathBuf> {
+    match env::var_os("_ZO_EXCLUDE_DIRS") {
+        Some(dirs_osstr) => env::split_paths(&dirs_osstr).collect(),
+        None => Vec::new(),
+    }
 }
 
 pub fn zo_maxage() -> Result<Rank> {
--- a/src/subcommand/add.rs
+++ b/src/subcommand/add.rs
@@ -2,13 +2,15 @@ use crate::config;
 use crate::util;
 
 use anyhow::{Context, Result};
-use std::env;
 use structopt::StructOpt;
+
+use std::env;
+use std::path::PathBuf;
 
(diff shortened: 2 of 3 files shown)


src/config.rs
```diff
@@ -39,3 +39,10 @@ pub fn zo_maxage() -> Result<Rank> {
         None => Ok(1000.0),
     }
 }
+
+pub fn zo_exclude_dirs() -> Option<Vec<PathBuf>> {
```
ajeetdsouza: Rather than `Option<Vec<PathBuf>>`, why not return an empty `Vec` if the environment variable is not set? It simplifies the code, and shouldn't have much of an impact on performance.

Also, a minor nitpick: since there isn't a logical way to order the functions in this module, could you order them alphabetically instead? I think it would make things easier once the number of environment variables increases.

src/subcommand/add.rs
```diff
@@ -16,13 +17,28 @@ impl Add {
         let mut db = util::get_db()?;
         let now = util::get_current_time()?;
         let maxage = config::zo_maxage()?;
+        let exclude_dirs = config::zo_exclude_dirs();
 
         match &self.path {
```
ajeetdsouza: To avoid code duplication, I think it would be a lot better if we did something like this:

```rust
let path = match &self.path {
    Some(path) => path.clone(),
    None => env::current_dir().context("unable to fetch current directory")?,
};
```

Also, do change the type of `Add::path` to `PathBuf`.

src/config.rs
cole-h: Roger. I'll do the same for the README, if you don't mind.

src/config.rs
```diff
@@ -1,10 +1,11 @@
-use crate::types::Rank;
```
ajeetdsouza: There doesn't seem to be much documentation on what the correct import order is supposed to be, but `rustfmt` does seem to want to put `crate` imports on top. I had a look at the official Rust repo, and they seem to be (mostly) going with:

1. `crate` imports
2. external imports
3. `std` imports

Perhaps we could go with that instead?

cole-h: Thanks for doing the legwork on researching the "standardized" (if only by popularity) order. I originally used what I do in all my projects, but as long as they're separated in some way, I don't mind which order it ends up being.

ajeetdsouza:
Thank you for your work on this!

PR #50, opened 2020-03-31
Show commit hash in version when built from source

cole-h:
Users might file an issue while running an older version, but the issue
has already been fixed in master. Adding the git revision to the version
output will expedite this diagnosis. For example:

    $ zoxide -V
    zoxide 0.3.0-0191eea

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Commit hash to version output.
+- Version output displays `git` revision information.
 
 ## [0.3.0] - 2020-03-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Commit hash to version output.
+
 ## [0.3.0] - 2020-03-30
 
 ### Added
--- a/build.rs
+++ b/build.rs
@@ -1,14 +1,35 @@
+use std::process::Command;
+
 fn main() {
-    let mut hash = std::process::Command::new("git")
-        .args(&["rev-parse", "--short", "HEAD"])
-        .output()
+    let status_code = Command::new("git")
+        .args(&["describe", "--tags", "--exact-match", "--dirty"])
+        .status()
         .ok()
-        .and_then(|proc| String::from_utf8(proc.stdout).ok())
-        .unwrap_or_default();
+        .and_then(|status| status.code());
 
-    if !hash.is_empty() {
-        hash = format!("-{}", hash);
-    }
+    let is_tagged = match status_code {
+        Some(code) => code == 0,
+        None => false,
+    };
 
-    println!("cargo:rustc-env=GIT_HASH={}", hash);
+    // If this is a tagged commit (a release), we don't want to include the
+    // commit hash in the version output.
+    let revision = if is_tagged {
+        String::new()
+    } else {
+        let mut hash = Command::new("git")
+            .args(&["rev-parse", "--short", "HEAD"])
+            .output()
+            .ok()
+            .and_then(|proc| String::from_utf8(proc.stdout).ok())
+            .unwrap_or_default();
+
+        if !hash.is_empty() {
+            hash = format!("-{}", hash);
+        }
+
+        hash
+    };
+
+    println!("cargo:rustc-env=GIT_HASH={}", revision);
 }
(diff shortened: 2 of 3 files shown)


ajeetdsouza:
This is a great idea! However, this would show commit hashes even on tagged releases, which I don't think is ideal. 

I will look into this further once I have more time. In the meantime, do take a look at [hasura/graphql-engine's version script](https://github.com/hasura/graphql-engine/blob/master/scripts/get-version.sh) for some inspiration, I think it's done quite well.

cole-h:
Thanks for that link! I was able to implement a check based on whether or not the current commit is tagged using the command from their `GIT_TAG_EXACT`. If the current commit *is* tagged (the `GIT_TAG_EXACT` command returned a zero exit code), the hash will not be used. If it's not tagged (the `GIT_TAG_EXACT` command returned a non-zero exit code), we append the hash.

ajeetdsouza:
Here's my take:

Instead of building our own version numbers, we can take them directly from Git:

```rust
git describe --tags --broken
```

- It tells us if the user has modified the repository locally
- We can copy-paste the version number and `git checkout` into it
- It's fairly readable, and doesn't add extra information if the commit is tagged

An environment variable name of `ZOXIDE_VERSION` would probably be more appropriate here.

In this case, `build.rs` would look something like:

```rust
if status_command == Some(0) { // command succeeded
    println!("cargo:rustc-env=ZOXIDE_VERSION={}", todo!());
}
```

And then in `main.rs`, we'd use something like:

```rust
option_env!("ZOXIDE_VERSION").unwrap_or(concat!("v", env!("CARGO_PKG_VERSION"), "-unknown"))
```

cole-h:
I opted to not go the `option_env!`-in-`main.rs` route (because I feel it would just look ugly) and instead put all the version-formatting logic in `build.rs`. I personally don't like seeing how many commits ahead the current version is, like `git describe` does (`tag-ahead-ghash`), but I'll follow your lead here. (I think it looks ugly, but that's probably just me.)

build.rs
```diff
@@ -0,0 +1,17 @@
+fn main() {
+    let git_describe = std::process::Command::new("git")
+        .args(&["describe", "--tags", "--broken"])
+        .output()
+        .ok()
+        .and_then(|proc| String::from_utf8(proc.stdout).ok());
+
+    let mut version_info = format!("v{}-unknown", env!("CARGO_PKG_VERSION"));
```
cole-h: I did it this way because otherwise, there'd be two `println!`s (out of three total) with the same contents: one in the case that `description` is empty (e.g. the command failed -- it prints to stderr and has an empty stdout) and one in the case that `git` was not found (or any other reason that would cause the process or its stdout to be `None`).

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Commit hash to version output.
```
cole-h: Now that we show `tag-ahead-ghash` in the version output, this doesn't seem as accurate anymore. What do you think about something like:

```suggestion
- Version output now displays `git` revision information.
```

build.rs
ajeetdsouza: You could probably do this instead:

```rust
let version_info = match git_describe {
    Some(description) if !description.is_empty() => description,
    _ => format!("v{}-unknown", env!("CARGO_PKG_VERSION"));,
}
```

CHANGELOG.md
ajeetdsouza: Sounds good to me!

build.rs
cole-h: Hah, I keep forgetting you can do that stuff inside `match` expressions. Thanks for reminding me.

CHANGELOG.md
cole-h: (I dropped "now" because it's under the `Added` header -- it's a given this didn't happen before.)

ajeetdsouza:
> I personally don't like seeing how many commits ahead the current version is, like `git describe` does (`tag-ahead-ghash`)

FWIW, I don't much like it either, and there doesn't seem to be a flag to disable it. I would be open to changing this to just `tag-ghash` (one can `git checkout` using this format too), but I really like the fact that `git describe` offers all this in a single command:

- checks for broken/dirty repos
- doesn't add extra information when there is an exact match

build.rs
ajeetdsouza: FYI, if there's only one condition to match, you can also use the `matches!` macro as an if condition!

ajeetdsouza:
Merging this, thank you!

cole-h:
I don't see any better solutions, unfortunately -- at least none that can be achieved with a single `git` invocation. If either of us finds a better solution, we can fix it and add it to `Changed` down the line :)

PR #56, opened 2020-04-05
Fix cd with path that contains spaces with bash

BestPig:
Quote around $@ are not used correctly.

This quote is finished the quotation, not starting a new one--> "$@

So when a path like "This is a test" is passed to zoxide, what happens is
`zoxide query This is a test`
instead of
`zoxide query "This is a test"`

Quoting the whole things `$(cmd)` is not necessary, so just remove those quote fix the issue.

### DIFF
--- a/src/subcommand/init.rs
+++ b/src/subcommand/init.rs
@@ -201,7 +201,7 @@ _z_cd() {{
             return 1
         fi
     else
-        result="$(zoxide query "$@")" || return "$?"
+        result=$(zoxide query "$@") || return "$?"
         if [ -d "$result" ]; then
             _z_cd "$result" || return "$?"
         elif [ -n "$result" ]; then


ajeetdsouza:
Could you provide an example of an error that the current implementation might cause? I added a debug statement to the current implementation to verify this:

```
$ z "this is a test"                                                                                                                1 ↵ 
[src/main.rs:21] std::env::args_os() = ArgsOs {
    inner: [
        "zoxide",
        "query",
        "this is a test",
    ],
}
Error: no match found

$ z this is a test                                                                                                                  1 ↵ 
[src/main.rs:21] std::env::args_os() = ArgsOs {
    inner: [
        "zoxide",
        "query",
        "this",
        "is",
        "a",
        "test",
    ],
}
Error: no match found
```

This seems to be expected behavior.

BestPig:
Hmm, I don't understand, I cannot reproduce the bug :O.
I did it because I was bored to cannot go into a directory with space, but maybe the problem was elsewhere.
I alias z command to cd, but I changed my alias when I do some test with zoxide, maybe I has done badly.

ajeetdsouza:
Alright, no problem!

PR #60, opened 2020-04-08
Add support for configurable extra fzf arguments

lzybkr:
via the environment variable _ZO_FZF_EXTRA_ARGS

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ NOTE: PWD hooks are currently not supported for POSIX shells.
 - `$_ZO_EXCLUDE_DIRS`: list of directories separated by platform-specific
   characters ("`:`" on Linux and macOS, and "`;`" on Windows) to be excluded from
   the database
-- `$_ZO_FZF_ARGS`: extra arguments to pass to fzf, e.g. `--height 25`
+- `$_ZO_FZF_ARGS`: extra arguments to pass to `fzf`
 - `$_ZO_MAXAGE`: sets the maximum total rank after which entries start getting deleted
 
 [`dirs` documentation]: https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ NOTE: PWD hooks are currently not supported for POSIX shells.
 - `$_ZO_EXCLUDE_DIRS`: list of directories separated by platform-specific
   characters ("`:`" on Linux and macOS, and "`;`" on Windows) to be excluded from
   the database
-- `$_ZO_FZF_ARGS`: extra arguments to pass to `fzf`
+- `$_ZO_FZF_OPTS`: extra arguments to pass to `fzf`
 - `$_ZO_MAXAGE`: sets the maximum total rank after which entries start getting deleted
 
 [`dirs` documentation]: https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html
--- a/README.md
+++ b/README.md
@@ -128,8 +128,8 @@ NOTE: PWD hooks are currently not supported for POSIX shells.
 - `$_ZO_EXCLUDE_DIRS`: list of directories separated by platform-specific
   characters ("`:`" on Linux and macOS, and "`;`" on Windows) to be excluded from
   the database
+- `$_ZO_FZF_ARGS`: extra arguments to pass to fzf, e.g. `--height 25`
 - `$_ZO_MAXAGE`: sets the maximum total rank after which entries start getting deleted
-- `$_ZO_FZF_EXTRA_ARGS`: extra arguments to pass to fzf, e.g. `--height 25`
 
 [`dirs` documentation]: https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html
 
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ NOTE: PWD hooks are currently not supported for POSIX shells.
   characters ("`:`" on Linux and macOS, and "`;`" on Windows) to be excluded from
   the database
 - `$_ZO_MAXAGE`: sets the maximum total rank after which entries start getting deleted
+- `$_ZO_FZF_EXTRA_ARGS`: extra arguments to pass to fzf, e.g. `--height 25`
 
 [`dirs` documentation]: https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,7 @@ use crate::db::DBVersion;
 use crate::dir::Rank;
 
 use anyhow::{bail, Context, Result};
+use shlex;
 
 use std::env;
 use std::fs;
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -129,6 +129,7 @@ NOTE: PWD hooks are currently not supported for POSIX shells.
   characters ("`:`" on Linux and macOS, and "`;`" on Windows) to be excluded from
   the database
 - `$_ZO_MAXAGE`: sets the maximum total rank after which entries start getting deleted
+- `$_ZO_FZF_EXTRA_ARGS`: extra arguments to pass to fzf, e.g. `--height 25`
```
ajeetdsouza, line 132: Do arrange the environment variables in alphabetical order, to make the list easier to scan through!

src/config.rs
```diff
@@ -51,3 +52,19 @@ pub fn zo_maxage() -> Result<Rank> {
         None => Ok(1000.0),
     }
 }
+
+pub fn zo_fzf_extra_args() -> Result<Vec<String>> {
```
ajeetdsouza, line 56: Here, too, it would be great if you could arrange the functions in alphabetical order, since there is no other way to logically arrange the functions, and would prove helpful as the codebase grows!

src/util.rs
```diff
@@ -76,6 +76,7 @@ where
 {
     let mut fzf = Command::new("fzf")
         .args(&["-n2..", "--no-sort"])
+        .args(config::zo_fzf_extra_args()?)
```
ajeetdsouza, line 79: Assigning this to a variable first would improve readability.

src/config.rs
```diff
@@ -51,3 +52,19 @@ pub fn zo_maxage() -> Result<Rank> {
...
+    match env::var_os("_ZO_FZF_EXTRA_ARGS") {
```
ajeetdsouza, line 57: Something like this would save you from the layers of nesting:

```rust
match env::var("_ZO_FZF_EXTRA_ARGS") {
    Ok(fzf_args_str) => {
        shlex::split(fzf_args_str).ok_or_else(anyhow!("could not parse _ZO_FZF_EXTRA_ARGS"))
    }
    Ok(env::VarError::NotPresent) => Ok(Vec::new()),
    e @ Err(env::VarError::NotUnicode) => e.context("invalid Unicode in _ZO_FZF_EXTRA_ARGS"),
}
```

src/config.rs
```diff
@@ -2,6 +2,7 @@ use crate::db::DBVersion;
 use crate::dir::Rank;
 
 use anyhow::{bail, Context, Result};
+use shlex;
```
cole-h, line 5:
```suggestion
```

cole-h:
Somewhat off-topic for this PR, but related to @ajeetdsouza's earlier comment on configuration:

While TOML/YAML configuration might be nice, I think that env var configuration is nicer. Want to test something? `_ZO_DATA_DIR=/tmp/zo zoxide add /tmp`. Want to change the max age for a single shell session? `export _ZO_MAXAGE=9999` (or however you achieve that in any arbitrary shell).

If we ship a default configuration file, we then have to make sure users know when new options are added, or options are deleted/deprecated. If we don't ship a default configuration, we have to make sure users know how to create their own. It also becomes another folder related to `zoxide` on users' filesystems: their local data dir for the database itself, and now the config dir for the configuration file. The file format itself also brings with it a whole host of issues, see: the many different "bugs" filed on Alacritty's issue tracker about configuration -- improper indentation, duplicate keys, unused keys, etc., in their case of YAML.

Overall, I much prefer having to use splitting logic over a separate config file, even if env vars were never truly removed.

(If there's still more to discuss on this topic, maybe it would be better to open an issue and ping me there, or even email me, to prevent cluttering this PR further.)

README.md
```diff
@@ -128,6 +128,7 @@ NOTE: PWD hooks are currently not supported for POSIX shells.
 - `$_ZO_EXCLUDE_DIRS`: list of directories separated by platform-specific
   characters ("`:`" on Linux and macOS, and "`;`" on Windows) to be excluded from
   the database
+- `$_ZO_FZF_ARGS`: extra arguments to pass to fzf, e.g. `--height 25`
```
cole-h, line 131:
```suggestion
- `$_ZO_FZF_ARGS`: extra arguments to pass to `fzf`
```

We don't have examples for any of the other env vars, so I don't think we need one here. Also, backtick-quote `fzf` because it's a command ;)

lzybkr:
OK, hopefully this is ready now.

It might be nice to add a suggested setting for $_ZO_FZF_ARGS in the README.md but I haven't settled on a "perfect" suggestion yet. So far I'm liking:

```powershell
$env:_ZO_FZF_ARGS = "--height 15% --layout=reverse --info=hidden --preview ""ls {2}"""
```

cole-h:
If we want to suggest anything at all, I'd rather go for something simple: `_ZO_FZF_ARGS="--height 20%"` (or `15%` as you have it). I don't like the idea of suggesting something as complicated as yours -- if a user wants to tweak further, they should do research into what options/flags `fzf` supports.

ajeetdsouza:
I think it would be best to add a link to the fzf documentation.

I also just checked the `fzf` docs, and it appears we have a solution to our splitting issue! Apparently, `fzf` uses `$FZF_DEFAULT_OPTS`, which we can use for our implementation.

I suggest:
- We change the variable name to `$_ZO_FZF_OPTS` for consistency.
- We use the `Command::env` function to set `$FZF_DEFAULT_OPTS` when starting `fzf`. This will also ensure that we do not clash with the user's `$FZF_DEFAULT_OPTS`, which they may be using for a different purpose.

lzybkr:
I considered using `$FZF_DEFAULT_OPTS` but decided against that. The scenarios I considered:

* User is using `$FZF_DEFAULT_OPTS` for UI settings - they probably wouldn't use `$_ZO_FZF_OPTS`.
* User is using `$FZF_DEFAULT_OPTS` for  other settings, e.g. history - we probably shouldn't override their settings.

ajeetdsouza:
I can think of 2 solutions to this:

1. `_ZO_FZF_OPTS` always overrides `FZF_DEFAULT_OPTS`. This means that an unset `_ZO_FZF_OPTS` will run `fzf` without any extra configuration. This keeps the behaviour for zoxide entirely separate, and if the user wants to keep the configuration the same, he can set `_ZO_FZF_OPTS="$FZF_DEFAULT_OPTS"`.

2. `_ZO_FZF_OPTS` overrides `FZF_DEFAULT_OPTS` only when it's set. This is less consistent, but it also means that `fzf` will work for the user out of the box as expected.

Of course, in either case, if the user wants default functionality and something extra, they could always go with `_ZO_FZF_OPTS="$FZF_DEFAULT_OPTS --height 15%"`.

lzybkr:
Having given this more thought, I think this PR isn't that useful.

If one wants to specify custom options, they should simply replace the appropriate aliases (`zi`, `zqi`, and `zri`) with their own alias/function/script (depending on their shell and intentions) that defines their preferred options, e.g. a bash user could use:

```bash
alias zi='FZF_DEFAULT_OPTS="--height 20" z -i'
```

ajeetdsouza:
That is true. One could also alias `zoxide` itself to change the interactive mode of both `zoxide query` and `zoxide remove`. I'll close this PR.

PR #62, opened 2020-04-09
Remove Windows verbatim drive in path

lzybkr:
`canonicalize` replaces the drive in a path with verbatim syntax,
e.g. `C:\Windows` turns into `\\?\C:\Windows`.

While Win32 apis handle verbatim paths nicely, commonly used Windows
shells like PowerShell and cmd do not, so we remove the prefix
before adding any paths to the database.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ anyhow = "1.0.28"
 bincode = "1.2.1"
 clap = "2.33.0"
 dirs = "2.0.2"
+dunce = "1.0"
 serde = { version = "1.0.106", features = ["derive"] }
 structopt = "0.3.12"
 uuid = { version = "0.8.1", features = ["v4"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ anyhow = "1.0.28"
 bincode = "1.2.1"
 clap = "2.33.0"
 dirs = "2.0.2"
-dunce = "1.0"
+dunce = "1.0.0"
 serde = { version = "1.0.106", features = ["derive"] }
 structopt = "0.3.12"
 uuid = { version = "0.8.1", features = ["v4"] }
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,6 +1,5 @@
 use crate::config;
 use crate::dir::{Dir, Epoch, Rank};
-use crate::util;
 
 use anyhow::{anyhow, bail, Context, Result};
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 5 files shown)


cole-h:
Maybe instead of having to loop through every path component, we could use the [`dunce` library](https://docs.rs/dunce/1.0.0/dunce/) for this? It seems ready-made for this situation.

Note: even though its `canonicalize` function is the same as `fs::canonicalize` on Unix systems, I think it would be better to only pull in `dunce` as a dependency on Windows.

ajeetdsouza:
`dunce` seems like a good idea. The conditional import may not be necessary, since [dunce does this anyway](https://docs.rs/dunce/1.0.0/src/dunce/lib.rs.html#59).

The only downside here would be the overhead of pulling in another crate - I tried compiling with `dunce` on Linux and the binary size _reduced_, oddly enough! On the plus side, it does reduce the amount of conditional compilation within the crate.

cole-h:
I was more worried about pulling in yet another dependency when unnecessary, considering my previous anti-bloat comments. However, thinking on it a bit more, I don't see a reason why we wouldn't use `dunce::canonicalize` on all platforms -- would at least keep code clearer than my idea of `#cfg[windows]` and `#cfg[not(windows)]`.

ajeetdsouza:
Also @lzybkr, would you be able to test the results on Windows? I don't have a Windows machine on me at the moment, so I can't test this myself.

lzybkr:
@cole-h - thanks for mentioning `dunce`, I'm actually using it in another project but had long forgotten about it.

@ajeetdsouza - I am testing on Windows (including the fzf change). I'll have PowerShell support in another PR after I'm happy with the PowerShell code that needs to be added.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ anyhow = "1.0.28"
 bincode = "1.2.1"
 clap = "2.33.0"
 dirs = "2.0.2"
+dunce = "1.0"
```
cole-h, line 20:
```suggestion
dunce = "1.0.0"
```

Cargo dependencies are a little weird. It's been a while since I had this explained to me, but I think it went something along the line of "1.0 will allow 1.0 and 1.1, while 1.0.0 will allow 1.0.0-1.0.99999...". Since 1.0 -> 1.1 is allowed to be a breaking change under semver, we don't want to allow that.

Even if that's no longer the case (or I'm plain wrong), doesn't hurt to be consistent with the rest of the file.

src/db.rs
```diff
@@ -144,7 +144,7 @@ impl DB {
                             continue;
                         }
                     };
-                    let path_abs = match Path::new(path_str).canonicalize() {
+                    let path_abs = match dunce::canonicalize(Path::new(path_str)) {
```
ajeetdsouza, line 147: The `str` type implements `AsRef<Path>` (which `dunce::canonicalize` uses as an argument), so `Path::new` is unnecessary here.

Cargo.toml
ajeetdsouza: The [Rust docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) mention how version numbers are handled. I've gone with what seems to be convention.

ajeetdsouza:
This is perfect, thank you!

PR #65, opened 2020-04-11
Add support for PowerShell

lzybkr:


src/subcommand/init.rs
```diff
@@ -215,6 +226,38 @@ _z_cd() {{
...
+
+    if ($args.Length -eq 0) {{
+        z_cd ~
+    }}
+    elseif ($args.Length -eq 1 -and $args[0] -eq '-') {{
+        z_cd -
+    }}
+    else {{
+        $result = zoxide query @args
+        if ($LASTEXITCODE -eq 0) {{
```
ajeetdsouza, line 251: `zoxide query --help` also returns an exit code of `0`, but we wouldn't want to `cd` in that case. Could you change this to `cd` only when `$result` is a valid directory, and, in the `else` clause, print out `$result`?

README.md
```diff
@@ -109,6 +110,17 @@ displayed.
 
 NOTE: PWD hooks are currently not supported for POSIX shells.
 
+#### PowerShell
+
+Add the following line to your profile:
+
+```powershell
+Invoke-Expression (& {
+    $hook = if ($PSVersionTable.PSVersion.Major -lt 6) { 'prompt' } else { 'pwd' }
```
ajeetdsouza, line 119: Why not always default to `prompt`? This will ensure that the PowerShell implementation stays consistent with the other shells.

src/subcommand/init.rs
```diff
@@ -215,6 +226,38 @@ _z_cd() {{
...
 use posix_z as zsh_z;
 
+fn powershell_z(z_cmd: &str) -> String {
+    format!(
+        r#"
+function {} {{
+    function z_cd($dir) {{
+        try {{
+            Set-Location $dir -ea Stop
+            if ($env:_ZO_ECHO -eq "1") {{
```
ajeetdsouza, line 236: Currently, the other shells simply check if `_ZO_ECHO` is set to a non-empty string, rather than a specific value like `1`. It would be great if PowerShell also matches this behaviour.

README.md
lzybkr: I don't think consistency should matter if there is a better way, and running `zoxide` less frequently seems better to me. I also think few know about the `pwd` hook and it would go unused otherwise.

Also - placement of this snippet matters more for the `prompt` hook as users typically define their own prompt but do not use the `pwd` hook. This snippet relies on being placed after a previously defined `prompt` function.

cole-h: We document the hooks we support [in the README](https://github.com/ajeetdsouza/zoxide#init-flags). Considering the section documenting this is right below your new PowerShell documentation, I don't think this is a good justification.

Why does "running `zoxide` less frequently" sound better to you? Although I was the original proponent for a PWD hook, I haven't actually used it since it was implemented.

Placement of the snippet is not something we can control. That's why you see the `posix` hook adding to the user's `PS1`. We can't prevent the user from overwriting the hook.

If a user opts to use the `prompt` hook, then we still have the same problem where it must be placed after a previously defined `prompt` function.

lzybkr: I wasn't aware of the `pwd` hook in any shell until I had started implementing this PR. I didn't even read all of the README, so I expect plenty of people to miss that.

I also didn't realize PowerShell had this hook since it is new, so I think it will be easy to miss and easy for people to assume you must modify your prompt, which in turn will turn some people away.

As for running less frequently - I know it's fast, but on Windows, especially in corporate domains - sometimes even fast things can be slow. It might be an IT department logging that you can't disable, or anti-virus or whatever, so running less often is just wiser IMO.

src/subcommand/init.rs
lzybkr: I can change this, but I'm not a fan of the value `0` being treated equivalently. I do think it's common to test the value `1` though.

README.md
ajeetdsouza: I do see what you mean, and I myself have been using the pwd hook on my system ever since it was implemented. However, I'd still like to keep the behaviour uniform - simply because it is less surprising to the user.

Assuming we default to a pwd hook unless it is unavailable, I can think of an example where this might be a problem - say, a user uses both bash and a POSIX shell on his system and initializes zoxide on them. Since the POSIX shell does not have a pwd hook, we default to the prompt hook. In this case, entries created by the POSIX shell will completely overshadow the entries created by bash, due to the frequent updates. Another (minor) issue I can think of would be that interactive queries would show much greater frecencies on some systems, without much of an explanation.

Admittedly, these are contrived examples, but I hope you see my point. I think requiring the user to explicitly enable pwd hooks ensures that they know what they're doing, and prevents confusion.

I can think of two solutions to this:
1. We could add a line to the README stating that users who need performance should go for pwd hooks.
2. We could default to pwd hooks (breaking change), and fail if they are unavailable with a help message:

```
PWD hooks are currently not available for this shell. Consider adding `--hook prompt` to your `init` command to fix this.
```

src/subcommand/init.rs
ajeetdsouza: That's a fair point. I will change the behaviour of the other shells to match this, then.

README.md
ajeetdsouza: @lzybkr I've changed the default hook to PWD in https://github.com/ajeetdsouza/zoxide/commit/7eed1ed9c3ac3eec27896329a05d98ee3ee9d012. Could you change this line to use the PWD hook by default, but explicitly require `--hook prompt` if it is unavailable?

src/subcommand/init.rs
ajeetdsouza: Other shells now require `_ZO_ECHO` to be set to `1` too, so I'm marking this as resolved.

README.md
lzybkr: @ajeetdsouza - I believe this implements exactly what you're asking for. The `pwd` hook is only available in PowerShell versions 6 or greater.

src/subcommand/init.rs
lzybkr: @ajeetdsouza - Fixed

lzybkr:
@ajeetdsouza - Are there any other changes you'd like to see?

src/subcommand/init/shell/powershell.rs
```diff
@@ -0,0 +1,84 @@
...
+}}
+"#,
+        z_cmd
+    )
+}
+
+fn alias(z_cmd: &str) -> String {
+    format!(
+        r#"
+function zi {{ {} -i @args }}
```
ajeetdsouza, line 53: The previous commit changes the aliases when `cmd` is changed. For example, this alias would be written as:

```
function {}i {{ {} -i @args }}
```

src/subcommand/init/shell/powershell.rs
```diff
@@ -0,0 +1,84 @@
...
+pub const CONFIG: ShellConfig = ShellConfig {
+    z,
+    alias,
+    hook: HookConfig {
+        prompt: HOOK_PROMPT,
+        pwd: hook_pwd,
+    },
+};
+
+fn z(z_cmd: &str) -> String {
```
ajeetdsouza, line 16: `--z-cmd` has been renamed to simply `--cmd`. Could you rename the instances of `z_cmd` here?

src/subcommand/init/shell/powershell.rs
```diff
@@ -0,0 +1,84 @@
...
+    format!(
+        r#"
+function {} {{
+    function z_cd($dir) {{
+        try {{
+            Set-Location $dir -ea Stop
+            if ($env:_ZO_ECHO -eq "1") {{
+                Write-Host "$PWD"
+            }}
+        }} catch {{
```
ajeetdsouza, line 26: I didn't quite get this - why do we silently ignore exceptions here?

src/subcommand/init/shell/powershell.rs
```diff
@@ -0,0 +1,84 @@
...
+$PreZoxidePrompt = $function:prompt
+function prompt {
+    $null = zoxide add
+    & $PreZoxidePrompt
+}
+"#;
+
+const fn hook_pwd() -> Result<Cow<'static, str>> {
+    const HOOK_PWD: &str = r#"
+$ExecutionContext.InvokeCommand.LocationChangedAction = {
```
ajeetdsouza, line 77: Is there a way we can throw a more descriptive error here in case the user's version of PowerShell is too old and he accidentally uses `--hook pwd`?

src/subcommand/init/shell/powershell.rs
lzybkr: Fixed (code was copied from `fish.rs` so I updated that file as well.

src/subcommand/init/shell/powershell.rs
lzybkr: I forget why I added it, but it's probably not needed now with `Test-Path` guarding the call.

src/subcommand/init/shell/powershell.rs
lzybkr: Done.

ajeetdsouza:
This is perfect, thank you!

PR #70, opened 2020-05-03
Aliases respect the z-cmd arg

fsommar:
When changing the z-cmd to something like `j` instead, the aliases will
also use `j` as a prefix, e.g. `ji`, `ja`.

This is, in my opinion, less surprising than keeping the aliases
prefixed with `z` even after changing its command.


ajeetdsouza:
This is great, thank you!

fsommar:
Thanks! When do you plan your next release?

ajeetdsouza:
`v0.4.0` has been released!

PR #84, opened 2020-06-07
Increase the _ZO_MAXAGE default (closes #83)

### LINKED ISSUE #83: zoxide forgets too easily by default
I have some directories that I use once a week or so, but zoxide seems to forget about them very easily. I think the default is a bit too conservative, could it be changed to remember many more items, if this wouldn't slow things down much?

skorokithakis:
This closes #83.


ajeetdsouza:
Thanks!

Issue #80, opened 2020-06-01
Zoxide stores the symlink targets


aruhier:
Hi,
Giving this example, with one symlink to another directory:

```
$ ls ~/tmp
  foo -> foo-target
  foo-target
```

If I cd into `~/tmp/foo`, zoxide will follow the symlink and stores this target (`foo-target`) in its database. So if I do:

```
$ zq foo
/home/anthony/tmp/foo-target
```

Which is a bit non-intuitive for me. Do you see a possibility of adding an option to ignore symlinks, and consider them as normal folders?

aruhier:
@ajeetdsouza: FYI, I'm working on this feature and will do a PR soon.

I was just a bit disappointed that `std::fs` doesn't have a function that gives the absolute path of a target without resolving the symlinks, so I had to quickly workaround it.

aruhier:
> I came up with the following (admittedly hacky) ways of solving this:
> 
>     * On Windows, `env::set_current_dir()` followed by `env::current_dir()` seems to return a normalized path without resolving symlinks, and fails if the path isn't valid - perfect for our use case.
> 
>     * However, `env::current_dir()` uses `getcwd()` under the hood, which on POSIX, is required to resolve symlinks. The upside here is that most shells also set the `PWD` environment variable, which does not resolve symlinks. So, we can validate `PWD` as an absolute path and fallback to `env::current_dir()` if the path is invalid. This is exactly what the [`pwd`](https://github.com/wertarbyte/coreutils/blob/master/src/pwd.c) utility does.

That's exactly what I wanted to do on my side. Here is what I wrote for the moment to get the `absolute_path` from a `base_dir`:

```rust
pub fn absolute_path<P: AsRef<Path>>(base_dir: &P, path: &P) -> Result<PathBuf> {
    let path = path.as_ref();
    let mut abs_path : PathBuf;

    // If path is an absolute path, base_dir is not needed.
    // "absolute" does not mean the path is clean (can be "/foo/../bar"), cleaning the path is still needed.
    if path.is_absolute() {
        abs_path = PathBuf::from(Component::RootDir.as_os_str())
    } else {
        abs_path = PathBuf::from(base_dir.as_ref());
    }

    for c in path.components() {
        match c {
            Component::CurDir | Component::RootDir => {},
            Component::ParentDir => {
                if abs_path.parent() != None {
                    abs_path.pop();
                }
            }
            _ => {
[...]

PR #85, opened 2020-06-08
Add _ZO_RESOLVE_SYMLINKS to resolve or not symlinks

aruhier:
Linked to #80.

_ZO_FOLLOW_SYMLINKS=0 disables the symlinks resolution, making a symlink and its target 2 different entries in the database.

**Example:**
* /tmp/foo-target is a directory
* /tmp/foo symlinks to /tmp/foo-target

With _ZO_FOLLOW_SYMLINKS=1 (default behavior), `z add /tmp/foo` adds `/tmp/foo-target` in the database.

With _ZO_FOLLOW_SYMLINKS=0, `z add /tmp/foo` adds `/tmp/foo` in the database.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -52,3 +52,10 @@ pub fn zo_maxage() -> Result<Rank> {
         None => Ok(10000.0),
     }
 }
+
+pub fn zo_resolve_symlinks() -> bool {
+    match env::var_os("_ZO_RESOLVE_SYMLINKS") {
+        Some(var) => var == "1",
+        None => false,
+    }
+}
--- a/src/subcommand/add.rs
+++ b/src/subcommand/add.rs
@@ -2,10 +2,9 @@ use crate::config;
 use crate::db::{Db, Dir, Rank};
 use crate::util;
 
-use anyhow::{Context, Result};
+use anyhow::Result;
 use structopt::StructOpt;
 
-use std::env;
 use std::path::{Path, PathBuf};
 
 /// Add a new directory or increment its rank
@@ -21,7 +20,7 @@ impl Add {
         let path = match &self.path {
             Some(path) => path,
             None => {
-                current_dir = env::current_dir().context("unable to fetch current directory")?;
+                current_dir = util::get_current_dir()?;
                 &current_dir
             }
         };
@@ -32,7 +31,11 @@ impl Add {
 
 fn add<P: AsRef<Path>>(path: P) -> Result<()> {
     let path = path.as_ref();
-    let path = util::canonicalize(&path)?;
+    let path = if config::zo_resolve_symlinks() {
+        util::canonicalize(&path)?
+    } else {
+        util::resolve_path(&path)?
+    };
 
     if config::zo_exclude_dirs().contains(&path) {
         return Ok(());
--- a/src/subcommand/import.rs
+++ b/src/subcommand/import.rs
@@ -1,3 +1,4 @@
+use crate::config;
 use crate::db::{Db, Dir};
 use crate::util;
 
@@ -39,7 +40,7 @@ fn import<P: AsRef<Path>>(path: P, merge: bool) -> Result<()> {
         .with_context(|| format!("could not read z database: {}", path.display()))?;
 
     for (idx, line) in buffer.lines().enumerate() {
-        if let Err(e) = import_line(&mut db, line) {
+        if let Err(e) = import_line(&mut db, line, config::zo_resolve_symlinks()) {
             let line_num = idx + 1;
             eprintln!("Error on line {}: {}", line_num, e);
         }
@@ -51,7 +52,7 @@ fn import<P: AsRef<Path>>(path: P, merge: bool) -> Result<()> {
     Ok(())
 }
 
-fn import_line(db: &mut Db, line: &str) -> Result<()> {
+fn import_line(db: &mut Db, line: &str, resolve_symlinks: bool) -> Result<()> {
     let mut split_line = line.rsplitn(3, '|');
 
     let (path, epoch_str, rank_str) = (|| {
@@ -70,7 +71,11 @@ fn import_line(db: &mut Db, line: &str) -> Result<()> {
         .parse::<f64>()
         .with_context(|| format!("invalid rank: {}", rank_str))?;
 
-    let path = util::canonicalize(&path)?;
+    let path = if resolve_symlinks {
+        util::canonicalize(&path)?
+    } else {
+        util::resolve_path(&path)?
+    };
     let path = util::path_to_str(&path)?;
 
     // If the path exists in the database, add the ranks and set the epoch to
--- a/README.md
+++ b/README.md
@@ -206,5 +206,6 @@ eval "$(zoxide init zsh)"
   ("`:`" on Linux/macOS, "`;`" on Windows) to be excluded from the database
 - `$_ZO_FZF_OPTS`: custom flags to pass to `fzf`
 - `$_ZO_MAXAGE`: sets the maximum total age after which entries start getting deleted
+- `$_ZO_RESOLVE_SYMLINKS`: when set to `1`, `z add` will resolve symlinks.
 
 [`dirs` documentation]: https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html
--- a/src/subcommand/init/shell/bash.rs
+++ b/src/subcommand/init/shell/bash.rs
@@ -15,7 +15,7 @@ pub const CONFIG: ShellConfig = ShellConfig {
 
 const HOOK_PROMPT: &str = r#"
 _zoxide_hook() {
-    zoxide add
+    zoxide add "$(pwd -L)"
 }
 
 case "$PROMPT_COMMAND" in
@@ -31,7 +31,7 @@ _zoxide_hook() {
         _ZO_PWD="${PWD}"
     elif [ "${_ZO_PWD}" != "${PWD}" ]; then
         _ZO_PWD="${PWD}"
-        zoxide add
+        zoxide add "$(pwd -L)"
     fi
 }
 
--- a/src/subcommand/init/shell/fish.rs
+++ b/src/subcommand/init/shell/fish.rs
@@ -69,14 +69,14 @@ end
 
 const HOOK_PROMPT: &str = r#"
 function _zoxide_hook --on-event fish_prompt
-    zoxide add
+    zoxide add $(pwd -L)
 end
 "#;
 
 const fn hook_pwd() -> Result<Cow<'static, str>> {
     const HOOK_PWD: &str = r#"
 function _zoxide_hook --on-variable PWD
-    zoxide add
+    zoxide add "$(pwd -L)"
 end
 "#;
 
--- a/src/subcommand/init/shell/posix.rs
+++ b/src/subcommand/init/shell/posix.rs
@@ -68,7 +68,7 @@ alias {0}r='zoxide remove'
 
 const HOOK_PROMPT: &str = r#"
 _zoxide_hook() {
-    zoxide add
+    zoxide add "$(pwd -L)"
 }
 
 case "$PS1" in
@@ -117,7 +117,7 @@ _zoxide_setpwd
 _zoxide_hook() {{
     _ZO_OLDPWD="$(cat "$_ZO_PWD_PATH")"
     if [ -z "$_ZO_OLDPWD" ] || [ "$_ZO_OLDPWD" != "$PWD" ]; then
-        _zoxide_setpwd && zoxide add > /dev/null
+        _zoxide_setpwd && zoxide add "$(pwd -L)" > /dev/null
     fi
 }}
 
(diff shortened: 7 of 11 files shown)


aruhier:
Sorry I totally missed this part from the linked issue:
```
So, we can validate PWD as an absolute path and fallback to env::current_dir() if the path is invalid. This is exactly what the pwd utility does.
```

I'll add a function for that in the `util` package.

ajeetdsouza:
> So, we can validate PWD as an absolute path and fallback to env::current_dir() if the path is invalid. This is exactly what the pwd utility does.

Unfortunately, this is only applicable in the case of calling `zoxide add` (without any parameters). Using `env::set_current_dir` will not modify `PWD`, as it comes from the parent shell.

The approach you have used looks fine to me, but I'll do some more research and get back to you ASAP.

aruhier:
True, but I'm also talking about this line for example: https://github.com/ajeetdsouza/zoxide/pull/85/files#diff-40620e05ae1c3f0b5171ef3415b2bdb6R38

Here if you give a relative path, you enter into the same case as `zoxide add` without parameter.

aruhier:
@ajeetdsouza: I added the `current_dir()` function in `util`.
I was able to test on linux, by unsetting `PWD`, setting it to an nonexisting directory. It seems to work as expected.

However, I'm not able to test on Windows or MacOS.

ajeetdsouza:
@aruhier 

---

I've looked this up, and I no longer think zoxide should handle `PWD`. There are two reasons for this:

- The `PWD` variable is a shell feature that isn't necessarily present on all shells. While every shell that zoxide _currently_ supports does have a `PWD` variable, we want to keep the core of zoxide shell-agnostic so that it can work on new shells seamlessly.

- The `PWD` variable is easily tampered with. There is a more correct way of going about this - `pwd` is a builtin on most shells, and can give us the current path without relying on the value of `PWD`. For a shell that does not implement this, it will fallback to GNU `pwd`, which will validate and print the value of `PWD` - similar to the behaviour you've implemented.

---

This is what I suggest we do:

- Change the hook functions in each shell to use the equivalent of `zoxide add "$(pwd -L)"` instead. For PowerShell, this would be `Get-Location`. When no path is specified, stick to `env::current_dir()` - this seems to be the only shell-agnostic option.

- When `_ZO_FOLLOW_SYMLINKS` is set, use `dunce::canonicalize` on the path. Otherwise, logically construct an absolute path and validate it manually, using `Path::is_dir`.

---

Two more notes:

- The `absolute_path()` function you wrote works great on Linux systems. There are a few issues with it on Windows. I'm currently reading about Windows paths and trying to add cross-platform functionality to your function, I'll let you know when it is complete.

- `dunce::canonicalize()` does its best to maximize compatibility of a path by removing the UNC prefix, but in cases where it cannot be removed, it returns them as is. This means that incompatible paths can be explicitly added to zoxide, rather than throwing an error. I think we should modify this function to have the same behaviour, but throw an error when the UNC prefix cannot be removed.

aruhier:
LGTM, thanks! I will change that soon.

src/config.rs
```diff
@@ -46,3 +46,16 @@ pub fn zo_maxage() -> Result<Rank> {
...
+        Some(follow_osstr) => {
+            let follow_str = follow_osstr.to_str();
+            match follow_str.unwrap_or("1").to_lowercase().as_ref() {
+                "0" | "false" | "no" => false,
+                _ => true,
+            }
+        }
+        None => true,
+    }
+}
```
ajeetdsouza, line 61: I think it would be much clearer to have environment variables default to the false value. Additionally, checking for multiple values brings additional complexity, and would not be consistent with the other environment variables.

I think we should change the name to something like `_ZO_RESOLVE_SYMLINKS`, which defaults to false unless explicitly set to `1`.

aruhier:
@ajeetdsouza: did you have some time to do a last review of my changes?

ajeetdsouza:
Hi @aruhier,

Sorry for the delay. I tried to extend your `absolute_path` function to work on Windows, but it seems Rust's `Path::components` is not very accurate on Windows. I'll work on this today and update you.

ajeetdsouza:
@aruhier, here's a cross-platform solution I wrote to resolve filepaths. `resolve_path` should remove `.` and `..` components, simplify all possible UNC paths while erroring out otherwise, and convert relative paths to absolute paths. Do let me know if you have suggestions, otherwise it would be nice if you could add it to your PR.

```rust
use anyhow::{bail, Context, Result};
use std::env;
use std::io::{self, stdin, BufRead, BufReader};
use std::path::{Component, Path, PathBuf};

fn get_current_dir() -> Result<PathBuf> {
    env::current_dir().context("could not get current path")
}

#[cfg(any(unix, windows))]
fn resolve_path<P: AsRef<Path>>(path: P) -> Result<PathBuf> {
    let path = path.as_ref();
    let base_path;

    let mut components = path.components().peekable();
    let mut stack = Vec::new();

    // initialize root
    if cfg!(unix) {
        match components.peek() {
            Some(Component::RootDir) => {
                let root = components.next().unwrap();
                stack.push(root);
            }
            _ => {
                base_path = get_current_dir()?;
                stack.extend(base_path.components());
            }
        }
    } else if cfg!(windows) {
        use std::path::Prefix;

        fn get_drive_letter<P: AsRef<Path>>(path: P) -> Option<u8> {
            let path = path.as_ref();
            let mut components = path.components();

            match components.next() {
                Some(Component::Prefix(prefix)) => match prefix.kind() {
                    Prefix::Disk(drive_letter) | Prefix::VerbatimDisk(drive_letter) => {
                        Some(drive_letter)
                    }
                    _ => None,
                },
                _ => None,
            }
        }

        fn get_drive_path(drive_letter: u8) -> PathBuf {
            format!(r"{}:\", drive_letter as char).into()
        }

        fn get_drive_relative(drive_letter: u8) -> Result<PathBuf> {
            let path = get_current_dir()?;
            if Some(drive_letter) == get_drive_letter(&path) {
                return Ok(path);
            }

            if let Some(path) = env::var_os(format!("={}:", drive_letter as char)) {
                return Ok(path.into());
            }

            let path = get_drive_path(drive_letter);
            Ok(path)
        }

        match components.peek() {
            Some(Component::Prefix(prefix)) => match prefix.kind() {
                Prefix::Disk(drive_letter) => {
                    let disk = components.next().unwrap();
                    match components.peek() {
                        Some(Component::RootDir) => {
                            let root = components.next().unwrap();
                            stack.push(disk);
                            stack.push(root);
                        }
                        _ => {
                            base_path = get_drive_relative(drive_letter)?;
                            stack.extend(base_path.components());
                        }
                    }
                }
                Prefix::VerbatimDisk(drive_letter) => {
                    components.next();
                    if components.peek() == Some(&Component::RootDir) {
                        components.next();
                    }

                    base_path = get_drive_path(drive_letter);
                    stack.extend(base_path.components());
                }
                _ => bail!("invalid path: {}", path.display()),
            },
            Some(Component::RootDir) => {
                components.next();

                let current_dir = env::current_dir()?;
                let drive_letter = get_drive_letter(&current_dir).with_context(|| {
                    format!("could not get drive letter: {}", current_dir.display())
                })?;
                base_path = get_drive_path(drive_letter);
                stack.extend(base_path.components());
            }
            _ => {
                base_path = get_current_dir()?;
                stack.extend(base_path.components());
            }
        }
    }

    for component in components {
        match component {
            Component::Normal(_) => stack.push(component),
            Component::CurDir => (),
            Component::ParentDir => {
                if stack.last() != Some(&Component::RootDir) {
                    stack.pop();
                }
            }
            Component::Prefix(_) | Component::RootDir => unreachable!(),
        }
    }

    let result = stack.iter().collect::<PathBuf>();
    if !result.is_dir() {
        bail!("could not resolve path: {}", result.display());
    }

    Ok(result)
}
```

src/config.rs
```diff
@@ -46,3 +46,16 @@ pub fn zo_maxage() -> Result<Rank> {
...
+    match env::var_os("_ZO_RESOLVE_SYMLINKS") {
+        Some(resolve_osstr) => {
+            let resolve_str = resolve_osstr.to_str();
+            match resolve_str.unwrap_or("0").to_lowercase().as_ref() {
+                "1" => true,
+                _ => false,
+            }
+        }
+        None => false,
+    }
```
ajeetdsouza, line 60: This can be simplified:

```rust
match env::var_os("_ZO_RESOLVE_SYMLINKS") {
    Some(var) => var == "1",
    None => false,
}
```

src/subcommand/add.rs
```diff
@@ -31,7 +31,14 @@ impl Add {
 
 fn add<P: AsRef<Path>>(path: P) -> Result<()> {
     let path = path.as_ref();
-    let path = canonicalize(&path)?;
+    let current_dir;
+    let path = match config::zo_resolve_symlinks() {
+        true => canonicalize(&path)?,
```
ajeetdsouza, line 36: if-else might be clearer than matching against boolean values.

cole-h:
Slight nit on the `resolve_path` suggestion: instead of runtime `cfg!()`s, I think it would probably be better to use a compiletime `#[cfg()]` (unless there is a reason you decided to use that instead).

ajeetdsouza:
@cole-h to the best of my knowledge, both variants are resolved at compile time; the only difference is:
- `#[cfg(..)]` removes the next block of code if the condition does not match. Thus, if the condition does not match, compilation will succeed regardless of any errors in that block.
- `cfg!(..)` gets replaced with `true`/`false` at compile time. Using it in an if condition means that it gets simplified to `if true`, and the optimizer removes the unnecessary branch. However, even if the condition does not match, the code is still being compiled before being optimized out, and must therefore be error-free.

I put it in an if condition since I get type-checking and IDE support for both target platforms, with no runtime cost (this is only possible because I'm use any OS-specific features). I also think it's a bit clearer, since it's a block in the middle of a function. That's just my opinion, though - I'd love to hear your thoughts.

aruhier:
> @cole-h to the best of my knowledge, both variants are resolved at compile time; the only difference is:
> 
>     * `#[cfg(..)]` removes the next block of code if the condition does not match. Thus, if the condition does not match, compilation will succeed regardless of any errors in that block.
> 
>     * `cfg!(..)` gets replaced with `true`/`false` at compile time. Using it in an if condition means that it gets simplified to `if true`, and the optimizer removes the unnecessary branch. However, even if the condition does not match, the code is still being compiled before being optimized out, and must therefore be error-free.
> 
> 
> I put it in an if condition since I get type-checking and IDE support for both target platforms, with no runtime cost (this is only possible because I'm use any OS-specific features). I also think it's a bit clearer, since it's a block in the middle of a function. That's just my opinion, though - I'd love to hear your thoughts.

Hmm so personally I'm not a big fan of defining specific functions for a certain os in scope, instead of defining them globally in the package with a clear conditional compilation condition. But as you wish, it's really nitpicking here.

Otherwise it looks good to me.

Just why do you go with a `Vec` here instead of a `Path` directly?

src/subcommand/add.rs
aruhier: Done, I didn't know you could use if/else for variable assignation.

src/config.rs
aruhier: Nice catch, I was pretty stupid on this one.

aruhier:
And I'll try to merge your change with mine tomorrow. At the beginning I tried to allow to have the `base_path` as argument so it's a bit more generic, but I understood that with windows it will be quite messy to keep that. Otherwise it should be quite straightforward, thanks for having looked at that!

ajeetdsouza:
> Just why do you go with a Vec here instead of a Path directly?

@aruhier I did try both approaches, but using a `Vec` was orders of magnitude faster, which makes sense because `Vec` is much better at what we're doing than `PathBuf` - for example, `Vec::pop()` would be a simple O(1) operation, but `PathBuf::pop()` would have to parse the path for separators and truncate it accordingly, which would be O(N).

aruhier:
Ah makes sense, I didn't look at the implementation of `PathBuf` and thought that the complete path was contained in a `Vec`.

aruhier:
Sorry I'm pretty late with this merge. I included your change, thanks a lot!

aruhier:
True, I've changed it.

For `remove`, I always used `resolve_path` instead of `canonicalize`, as in my opinion you want to remove an exact path, and it seemed weird to me to rely on the option for it.
Tell me what you think and I just do the exact same thing as in `import` and `add`.

src/subcommand/import.rs
```diff
@@ -70,7 +71,11 @@ fn import_line(db: &mut Db, line: &str) -> Result<()> {
         .parse::<f64>()
         .with_context(|| format!("invalid rank: {}", rank_str))?;
 
-    let path = util::canonicalize(&path)?;
+    let path = if config::zo_resolve_symlinks() {
```
ajeetdsouza, line 74: `import_line()` is run once for each line in the input. To avoid parsing `_ZO_RESOLVE_SYMLINKS` every time, I suggest you pass a bool `resolve_symlinks` from the `import()` function.

aruhier: Done, thanks, I didn't look how this function was used.

ajeetdsouza:
@aruhier, this is great. Thanks for your work on this, I appreciate it!

aruhier:
And thanks a lot for the work done on your side! I rarely code in rust, so your comments and what you did for the Windows paths were super handy for me.

PR #87, opened 2020-06-19
Cleanup temporary $result variables

mqnfred:
After we call `$(zoxide query $input)`, we store it in `$result` in all
shells. This `result` environment variables survives after the call.

To fix this undesirable behavior, we rename the temporary variable to
something less ubiquitous (`_ZO_RESULT`,) as well as unset it after the
`cd`/`echo` has been done.

For the record, I noticed this because I use `%~` in my zsh `PS1`, which
has a tendency to re-use set environment variables in the path, ie:

```
pete:~$ z code/misc/rustyline
pete:~/code/misc/rustyline$ z
pete:~$ z rusty
pete:~result$ _
```

As you can see, on the second cd, the prompt path is `result`. This
hides the prompt information. I have not tested the fish and powershell
implementations, as I do not have the capabilities to do so.

### DIFF
--- a/src/subcommand/init/shell/posix.rs
+++ b/src/subcommand/init/shell/posix.rs
@@ -37,12 +37,13 @@ _z_cd() {{
             return 1
         fi
     else
-        result="$(zoxide query "$@")" || return "$?"
-        if [ -d "$result" ]; then
-            _z_cd "$result" || return "$?"
-        elif [ -n "$result" ]; then
-            echo "$result"
+        _zoxide_result="$(zoxide query "$@")" || return "$?"
+        if [ -d "$_zoxide_result" ]; then
+            _z_cd "$_zoxide_result" || return "$?"
+        elif [ -n "$_zoxide_result" ]; then
+            echo "$_zoxide_result"
         fi
+        unset _zoxide_result
     fi
 }}
 "#,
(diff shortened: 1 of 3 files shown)


ajeetdsouza:
Hi, thanks for pointing this out!

The variables defined in the fish and PowerShell functions are local variables, and do not have this problem. POSIX, unfortunately, does not define a way to set local variables - although the `local` keyword is extremely common, we cannot rely on it always being present.

The solution you proposed does seem to be the best way out for POSIX functions, but it would be nice if you could rename it to `_zoxide_result` to avoid confusion, since we use `_ZO_*` for configuration.

mqnfred:
Hey Ajeet, thank you for the quick reply.

I have modified the pull-request, I took the time to rename all the variables to the same name across all platforms (for the sake of consistency) and removed the unset/clear in fish/powershell as they are not required.

Let me know if anything else is amiss!

src/subcommand/init/shell/posix.rs
```diff
@@ -37,12 +37,13 @@ _z_cd() {{
...
         fi
     else
-        result="$(zoxide query "$@")" || return "$?"
-        if [ -d "$result" ]; then
-            _z_cd "$result" || return "$?"
-        elif [ -n "$result" ]; then
-            echo "$result"
+        _zoxide_result="$(zoxide query "$@")" || return "$?"
+        if [ -d "$_zoxide_result" ]; then
+            _z_cd "$_zoxide_result" || (unset _zoxide_result && return "$?")
```
ajeetdsouza, line 42: Calling `unset` here will reset the exit code that we're returning (`$?`). I think it should be safe to not use `unset`, since the variable is "namespaced" under `_zoxide_`.

ajeetdsouza:
Thanks, this is great!

PR #88, opened 2020-06-23
README: add instructions on installing via MacPorts

herbygillot:
[`zoxide` is now in Macports](https://ports.macports.org/port/zoxide/summary).  This PR adds instructions on how would install in this way.


ajeetdsouza:
Thanks for maintaining zoxide!

PR #96, opened 2020-07-08
_zoxide_hook ignore $? when running, adding a small way around to get it

tomMoulard:
# Problem
In my prompt, I display the return value of the last command but  _zoxide_hook ignore it.

# Solution
I've added a small workaround to get the correct value and not the one of zoxide.

I mainly use Bash and not other shell. Therefore, I could not fully test on POSIX or zsh shell.
I also could not implement it for Fish.

Comment are welcome

### DIFF
--- a/src/subcommand/init/shell/bash.rs
+++ b/src/subcommand/init/shell/bash.rs
@@ -37,7 +37,7 @@ _zoxide_hook() {
 
 case "$PROMPT_COMMAND" in
     *_zoxide_hook*) ;;
-    *) PROMPT_COMMAND="_zoxide_hook${PROMPT_COMMAND:+;${PROMPT_COMMAND}}" ;;
+    *) PROMPT_COMMAND="${PROMPT_COMMAND:+${PROMPT_COMMAND};}_zoxide_hook" ;;
 esac
 "#;
 


ajeetdsouza:
Hi @tomMoulard, could you help me understand your exact issue? I've tried the following with both pwd and prompt hooks, and it does seem to work as expected:

```sh
$ true
$ echo $?
0
$ false
$ echo $?
1
```

tomMoulard:
It's kind of a complex situation to be in.
In short, I catch `$?` at the beginning of my `promp` function to display it in the `PS1`.
Something like this:
```bash
PROMPT_COMMAND=prompt
prompt(){
	RETVAL=$?
	PS1='${RETVAL} \$ '
}
```
Which gives:
```bash
0 $ set -x
++ _zoxide_hook
++ '[' -z /root ']'
++ '[' /root '!=' /root ']'
++ prompt
++ RETVAL=0
++ PS1='${RETVAL} \$ '
0 $ true
+ true
++ _zoxide_hook
++ '[' -z /root ']'
++ '[' /root '!=' /root ']'
++ prompt
++ RETVAL=0
++ PS1='${RETVAL} \$ '
0 $ false
+ false
++ _zoxide_hook
++ '[' -z /root ']'
++ '[' /root '!=' /root ']'
++ prompt
++ RETVAL=0
++ PS1='${RETVAL} \$ '
0 $
```
(Note that if `PS1` is set to `'$(echo $?) \$ '` zoxide work as intended)

And if i use my patch, i get:
```bash
0 $ true
+ true
++ _zoxide_hook
++ local exit_code=0
++ '[' -z /root ']'
++ '[' /root '!=' /root ']'
++ return 0
++ prompt
++ RETVAL=0
++ PS1='${RETVAL} \$ '
0 $ false
+ false
++ _zoxide_hook
++ local exit_code=1
++ '[' -z /root ']'
++ '[' /root '!=' /root ']'
++ return 1
++ prompt
++ RETVAL=1
++ PS1='${RETVAL} \$ '
1 $ 
```
Is that a better explanation ?

ajeetdsouza:
Hey @tomMoulard,

Sorry for the late reply. I was testing out your idea today, and I think it would be simpler to just add `_zoxide_hook` to the end of `PROMPT_COMMAND` as follows:

```shell
case "$PROMPT_COMMAND" in
    *_zoxide_hook*) ;;
    *) PROMPT_COMMAND="${PROMPT_COMMAND:+${PROMPT_COMMAND};}_zoxide_hook" ;;
esac
```

What do you think?

ajeetdsouza:
Hi @tomMoulard, did you get a chance to look into this?

tomMoulard:
Hi, yes sorry, I get to it asap

tomMoulard:
This solution seems good, I changed the PR to fit your proposition.

Thanks a lot :+1:

ajeetdsouza:
Looks great. Thanks for your contribution!

Issue #111, opened 2020-10-20
zoxide init generates bash completions for non-bash shells


lmartinez-mirror:
After the latest update I received a syntax error from my init script for zoxide.

Running `zoxide init fish` returns a bash completion script instead of fish completions. This also occurs for the POSIX init.

A diff between the output for a bash and fish init shows they're identical.

[zoxide-fish-init.txt](https://github.com/ajeetdsouza/zoxide/files/5410289/zoxide-fish-init.txt)
[zoxide-bash-init.txt](https://github.com/ajeetdsouza/zoxide/files/5410294/zoxide-bash-init.txt)

cole-h:
```
git bisect start
# good: [a454f7cbbf7c5e84f0cc23e9dd0c3c668938d4f5] Release 0.4.3
git bisect good a454f7cbbf7c5e84f0cc23e9dd0c3c668938d4f5
# bad: [3d432105e045ceb879f0dd0e66fb72db994a2e3b] Retry loop when persisting on Windows
git bisect bad 3d432105e045ceb879f0dd0e66fb72db994a2e3b
# good: [2eb02cbd761a87eabde812d7ce6254d0f78b1fe0] Update dependencies
git bisect good 2eb02cbd761a87eabde812d7ce6254d0f78b1fe0
# good: [fcac28eed0ccbc3ae5aefaa3fb2d9cb6b722ede6] Use tool-cache for cargo-udeps
git bisect good fcac28eed0ccbc3ae5aefaa3fb2d9cb6b722ede6
# bad: [aca2477b689712d752ac8478b0fdb7692598f39d] Split crates
git bisect bad aca2477b689712d752ac8478b0fdb7692598f39d
# good: [eaf63bcc69dfbe576053f0744824ece4045b7ffa] Fix clippy lints
git bisect good eaf63bcc69dfbe576053f0744824ece4045b7ffa
# first bad commit: [aca2477b689712d752ac8478b0fdb7692598f39d] Split crates
```

cole-h:
The actual problem comes from this span:

https://github.com/ajeetdsouza/zoxide/commit/aca2477b689712d752ac8478b0fdb7692598f39d#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR165-R168

cole-h:
Fixed in https://github.com/ajeetdsouza/zoxide/pull/112.

ajeetdsouza:
[...]

PR #112, opened 2020-10-20
fix init script generation

cole-h:
A copy-paste error caused the Fish, POSIX, and Powershell shells to
generate an incorrect init script.

---

Closes https://github.com/ajeetdsouza/zoxide/issues/111.


ajeetdsouza:
Thanks!

PR #125, opened 2020-11-23
Parameterize fuzzy finder

poliorcetics:
This parameterize the fuzzy finder in `zi`, allowing for both `fzf` or `skim` (binary: `sk`).

To keep a backward compatible behaviour, `fzf` stays as the default, even when `_ZO_FUZZY_FINDER_CMD` is not set.


ajeetdsouza:
That's a very detailed PR! Before I review it, though, I have to ask - what makes you choose `skim` over `fzf`?

poliorcetics:
I already use `skim` in `neovim`, that way I avoid having to install two tools that do pretty much the exact same thing.

Also, I use Rust a **lot** more than I use Go so using Rust tools is more natural for me, though that's a completely personal thing.

ajeetdsouza:
Support for skim has been discussed in previous issues, but from what I've seen, fzf is both more featureful and more performant. So, I'm still not convinced why you would want skim support:

> I already use skim in neovim, that way I avoid having to install two tools that do pretty much the exact same thing.

I can see why having redundant tools installed would be annoying, but the fzf binary is only 3 MB. Installing both of them wouldn't have any impact on your system.

> Also, I use Rust a lot more than I use Go so using Rust tools is more natural for me

Admittedly, Cargo does provide great UX, but this can't be the entire reason for using skim, right?

Merging this PR would be easy, but like everything else, it has to be maintained, and it adds more complexity for the user. There are plenty of fzf clones out there, so I'll try to rephrase my question. How does skim add value to zoxide? Does it do anything fzf can't? Does it do anything better than fzf?

poliorcetics:
If you don't want to maintain it you can close this, it's a perfectly reasonable explanation. I just like being able to configure my tools but that's certainly not a critical feature. 😄

ajeetdsouza:
Thanks for understanding. Btw, you don't really need to fork zoxide to use skim with it - you can simply `zoxide query -ls <keywords>` and pipe to skim.

poliorcetics:
> `zoxide query -ls <keywords>`

I didn't know this, thanks !

EvanCarroll:
> Support for skim has been discussed in previous issues, but from what I've seen, fzf is both more featureful and more performant. So, I'm still not convinced why you would want skim support:

Could you quantify/clarify either of these. The recent work in skim leads me to believe it is as fast and no less featureful. The reason why you would want it is because you're more comfortable with the ecosystem... and stuff like this,

https://bodhi.fedoraproject.org/updates/FEDORA-2022-30c5ed5625

I also don't trust anything from junegunn.

Issue #126, opened 2020-12-11
Zoxide in symbolic link path are not working correctly when tring to walk through parent paths.


fx-kirin:
I found a bug and I'd like to report it here.

```
mkdir /tmp/zoxide-test
mkdir /tmp/zoxide-test/symbolic-links/
mkdir /tmp/zoxide-test/symbolic-links/actual_folder
mkdir /tmp/zoxide-test/actual-directories/

mkdir /tmp/zoxide-test/actual-directories/folder1
mkdir /tmp/zoxide-test/actual-directories/folder2
ln -s /tmp/zoxide-test/actual-directories/folder1 /tmp/zoxide-test/symbolic-links

z /tmp/zoxide-test/symbolic-links/folder1

# Error will happen here.
z ../actual_folder
# $ z ../actual_folder
# Error: no match found

# But this works
cd ../actual_folder

# When tried to get candidates with Tab Key in `/tmp/zoxide-test/symbolic-links`. 
z ../   
# $ z ../
# >
# > actual-directories/
#   symbolic-links/

# This command shows an error but moves current directory
z ../folder2
# /tmp/zoxide-test/symbolic-links/folder1
# $ z ../folder2
# Error: could not resolve path: /tmp/zoxide-test/symbolic-links/folder2
# /tmp/zoxide-test/actual-directories/folder2
# $ 

```

I'm usgin version `0.5.0`. It seems `zoxide` must using the actual path of symbolic link.

ajeetdsouza:
This is a really interesting issue, thank you for the detailed report!

> `z ../actual_folder` 

Fixed in https://github.com/ajeetdsouza/zoxide/commit/18121e07f972a8f33cfe36d76dc7097880a68dc3.

> `z ../folder2`

This appears to be a bug in zsh itself. I've sent a report to zsh-workers@zsh.org, but I'm including a copy here for reference:

```zsh
function print_dir() { pwd -L }
chpwd_functions=(print_dir)
mkdir -p /tmp/chdir-test/symbolic-links/
mkdir -p /tmp/chdir-test/actual-directories/folder1
mkdir -p /tmp/chdir-test/actual-directories/folder2
ln -s /tmp/chdir-test/actual-directories/folder1 /tmp/chdir-test/symbolic-links

# This works fine:
cd /tmp/chdir-test/symbolic-links/folder1
# >> /tmp/chdir-test/symbolic-links/folder1

# This prints a path that does not exist (try it with cd!)
[...]

PR #127, opened 2020-12-11
Remove -d check when changing directories

ajeetdsouza:
Fixes https://github.com/ajeetdsouza/zoxide/issues/126.

**Status:**

- [x] bash: fixed
- [ ] fish
- [x] posix: fixed
- [x] powershell: already works
- [x] xonsh: works as expected (both cd and z don't work)
- [x] zsh: fixed

### DIFF
--- a/templates/zsh.txt
+++ b/templates/zsh.txt
@@ -59,7 +59,7 @@ function __zoxide_z() {
             echo "zoxide: \\$OLDPWD is not set"
             return 1
         fi
-    elif [ "$#" -eq 1 ] &&  [ -d "$1" ]; then
+    elif [ "$#" -eq 1 ] &&  cd "$1" &>{{ Opts::DEVNULL }}; then
         __zoxide_cd "$1"
     else
         local __zoxide_result
--- a/templates/zsh.txt
+++ b/templates/zsh.txt
@@ -59,9 +59,7 @@ function __zoxide_z() {
             echo "zoxide: \\$OLDPWD is not set"
             return 1
         fi
-    elif [ "$#" -eq 1 ] &&  cd "$1" &>{{ Opts::DEVNULL }}; then
-        __zoxide_cd "$1"
-    else
+    elif ! { [ "$#" -eq 1 ] && __zoxide_cd "$1" &>{{ Opts::DEVNULL }}; }; then
         local __zoxide_result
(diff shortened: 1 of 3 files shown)


templates/zsh.txt
```diff
@@ -59,9 +59,7 @@ function __zoxide_z() {
             echo "zoxide: \\$OLDPWD is not set"
             return 1
         fi
-    elif [ "$#" -eq 1 ] &&  cd "$1" &>{{ Opts::DEVNULL }}; then
-        __zoxide_cd "$1"
-    else
+    elif ! { [ "$#" -eq 1 ] && __zoxide_cd "$1" &>{{ Opts::DEVNULL }}; }; then
```
ajeetdsouza, line 62: This has a problem - it hides the output of `__zoxide_cd` as well as any other PWD hook that the user might have.

templates/zsh.txt
```diff
@@ -59,7 +59,9 @@ function __zoxide_z() {
             echo "zoxide: \\$OLDPWD is not set"
             return 1
         fi
-    elif ! { [ "$#" -eq 1 ] && __zoxide_cd "$1" &>{{ Opts::DEVNULL }}; }; then
+    elif [ "$#" -eq 1 ] && (cd "$1" &>{{ Opts::DEVNULL }}) then
```
ajeetdsouza, line 62: The cd hook here calls zoxide upon completion, so `zoxide add` ends up getting called twice.

ajeetdsouza:
Unfortunately, there seems to be no clean way for zoxide to check if cd is possible to virtual directories in the shell. Here are some approaches:

1. cd-ing first and checking if it succeeded: this works fine, but prints to stderr when it fails. We could suppress output, but that would also suppress any useful output from a PWD hook (such as `_ZO_ECHO`).
2. Invoking cd in a subshell and checking if it succeeded: this causes `zoxide add` to be run twice if PWD hooks are set. Another way is to use an actual shell to check, but the same problem occurs. We could get a clean shell using `env -i`, but that would mean that we can't access environment variables, thus losing PWD.
3. Non-shell programs don't use virtual directories, and therefore can't be used to determine cd-ability. For example, `env -C ../actual_folder pwd` complains that `../actual_folder` does not exist.

From the looks of it, all possible approaches end up having some unintentional side effects, so I'm going to close this PR.

PR #131, opened 2020-12-13
Add Alpine Linux installation instructions

jirutka:
Note: Alpine 3.13 has not been released yet, it should be till the end of this year.


ajeetdsouza:
I've changed this to Alpine Linux Edge. This can be edited once Alpine 3.13 is out.

Thank you for maintaining zoxide!

PR #133, opened 2020-12-21
Add export subcommand

kidonng:
This is a simple addition, and it could even be a standalone tool. But I agree with [@ahmedelgabri's concern](https://github.com/ajeetdsouza/zoxide/issues/11#issuecomment-598896508), this should belong to zoxide.

The `export` subcommand outputs the same format as most `z` implementations such as rupa/z:

```
/home/runner/work/fastmac/fastmac|1|1608511969
/home/runner/.local/share|3|1608511994
/home/runner/.nvm|3|1608511986
/home/runner/.ssh|1|1608511984
```

### DIFF
--- a/src/cmd/export.rs
+++ b/src/cmd/export.rs
@@ -1,0 +1,27 @@
+use super::Cmd;
+use crate::config;
+
+use crate::store::Store;
+use anyhow::{Context, Result};
+use clap::Clap;
+
+use std::io::{self, Write};
+
+/// Export entries from database
+#[derive(Clap, Debug)]
+pub struct Export {}
+
+impl Cmd for Export {
+    fn run(&self) -> Result<()> {
+        let data_dir = config::zo_data_dir()?;
+        let store = Store::open(&data_dir)?;
+
+        let mut handle = io::stdout();
+        for dir in store.dirs.iter() {
+            writeln!(handle, "{}|{}|{}", dir.path, dir.rank, dir.last_accessed)
+                .context("could not write to stdout")?;
+        }
+
+        Ok(())
+    }
+}
--- a/src/cmd/mod.rs
+++ b/src/cmd/mod.rs
@@ -1,4 +1,5 @@
 mod add;
+mod export;
 mod import;
 mod init;
 mod query;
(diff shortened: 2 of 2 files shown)


kidonng:
Uh, seems like the current code doesn't handle broken pipe very well.

```
❯ cargo run export | head
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/zoxide export`
/Users/kid/Downloads|78|1608511191
/var/folders/yn/2sq2fsms4k7c1ltt2_rj50bm0000gn/T|27|1607738459
/var/folders/yn/2sq2fsms4k7c1ltt2_rj50bm0000gn/T/test|16|1606713571
/Users/kid/.config/fish|63|1608512725
/Users/kid/.config/fish/dotfiles|5|1605406205
/Users/kid/.config/fish/dotfiles/completions|2|1605406106
/Users/kid|345|1608528702
/Users/kid/.ssh|1|1605404218
/Users/kid/.config|222|1608478072
/Users/kid/.config/dotfiles|13|1608206944
Error: could not write to stdout

Caused by:
    Broken pipe (os error 32)
```

AFAIK other Rust tools don't solve this very well, most of them just hide the error. Should we do the same?

ajeetdsouza:
Hey @kidonng,

I appreciate your effort, but, as mentioned in the issue you linked, I don't really see this is an important goal for zoxide at the moment. In general, I'd like to keep zoxide as lean as possible, and add things only if they are essential or are significant enhancements.

Shipping an `export` subcommand:
- adds [bloat](https://en.wikipedia.org/wiki/Feature_creep) to what should ideally be a very simple tool, for a feature that most people would (hopefully) never need to use
- adds another subcommand for the user to parse through every time they go through the `--help` page
- ties zoxide to the `z` format - which it may not always be compatible with

In future, I'd appreciate if you could create an issue to request a feature before working on it - that way, you'd be able to incorporate everyone's feedback, saving you time and effort.

I'm closing this PR for now - I've tried to ensure that users don't need to use multiple `z` programs by ensuring that zoxide is available on all major shells and platforms. If there's a specific functionality you find missing, feel free to create an issue for it, and we can discuss it there :smile:

PR #146, opened 2021-02-02
Specified external call to cd for fish

### LINKED ISSUE #145: Fish: Allow cd aliasing
Really cool plugin, just started using it with fish.

Ran into issue trying to alias `cd` to `z`, where zoxide tries to call my alias, which calls zoxide, which calls my alias, until stack overflow.

`zoxide init fish` outputs this internal command (among other things):
```
function __zoxide_cd
    cd $argv
    and commandline -f repaint
end
```

However, it should be:
```
function __zoxide_cd
    command cd $argv
    and commandline -f repaint
end
```
In fish, command  <cmd> makes sure to call out instead of calling functions.

I'm not familiar with zoxide's codebase, so I didn't know where to put the fix, otherwise I'd PR myself.

fennewald:
I lied. I will push a fix lol.
Closes #145

### DIFF
--- a/templates/fish.txt
+++ b/templates/fish.txt
@@ -10,15 +10,15 @@
 # pwd based on the value of _ZO_RESOLVE_SYMLINKS.
 function __zoxide_pwd
 {%- if resolve_symlinks %}
-    pwd -P
+    command pwd -P
 {%- else %}
-    pwd -L
+    command pwd -L
 {%- endif %}
 end
 
 # cd + custom logic based on the value of _ZO_ECHO.
 function __zoxide_cd
-    cd $argv
+    command cd $argv
     {%- if echo %}
     and __zoxide_pwd
     {%- endif %}


ajeetdsouza (changes requested):
Thanks! A few comments:

Use `builtin` wherever possible instead of `command` (`cd`, `pwd`), since the builtin shell implementation is more reliable than an external command. For example, `command cd` does not work on my Linux machine. `builtin --names` will give you the complete list of shell builtins.

There are a couple more places where you could add `builtin` to resolve ambiguity:
- `commandline`
- `functions`

Things like `set` and `and`, while also builtins, do not need the `builtin` keyword, since they are [reserved keywords](https://github.com/fish-shell/fish-shell/blob/cf4f8ae929f05f9a794bd329de3599918dc95aab/src/parser_keywords.cpp#L70) and cannot be redefined.

On a similar note, you could also replace `zoxide` with `command zoxide` to reduce ambiguity. The only thing that I don't think we can change is `abbr` - which is neither a `builtin` nor a reserved keyword.

nikelborm:
> Thanks! A few comments:
> 
> Use `builtin` wherever possible instead of `command` (`cd`, `pwd`), since the builtin shell implementation is more reliable than an external command. For example, `command cd` does not work on my Linux machine. `builtin --names` will give you the complete list of shell builtins.
> 
> There are a couple more places where you could add `builtin` to resolve ambiguity:
> 
> * `commandline`
> * `functions`
> 
> Things like `set` and `and`, while also builtins, do not need the `builtin` keyword, since they are [reserved keywords](https://github.com/fish-shell/fish-shell/blob/cf4f8ae929f05f9a794bd329de3599918dc95aab/src/parser_keywords.cpp#L70) and cannot be redefined.
> 
> On a similar note, you could also replace `zoxide` with `command zoxide` to reduce ambiguity. The only thing that I don't think we can change is `abbr` - which is neither a `builtin` nor a reserved keyword.

Hi, @ajeetdsouza I may disagree. I created `alias cd="z"`and now every cd hangs, because `z` calls `_z_cd` which calls `cd` not as `command z`, but as my alias and it results in infinite loop

Are there any middle ground we can search for it? because I think not only me just set an alias for `cd` to point to `z`

ajeetdsouza:
You can just initialize zoxide as `zoxide init fish --cmd=cd`. Or, you can define the alias in your fish config file, the line after you call `zoxide init`.

PR #166, opened 2021-03-30
Reference integration with nnn (autojump plugin)

kentprimrose:
Added nnn (autojump plugin) to list of third party integrations


ajeetdsouza:
Thank you!

PR #167, opened 2021-03-30
Correctly escape zoxide arguments in powershell

davidkna:
PowerShell escapes CLI arguments according to the `cmd.exe` rules which differ from the rust way of escaping them. For instance this create can issues when an arguments ends with a trailing backslash.
This PR changes this by wrapping the zoxide exe with a PowerShell function and which manually escapes the arguments or uses `ArgumentList` when available.
As a side-effect this also makes the zoxide hook support long paths (`MAX_PATH`+).
The implementation is taken from starship: https://github.com/starship/starship/blob/master/src/init/starship.ps1#L25-L63.

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -5,6 +5,44 @@
 # Utility functions for zoxide.
 #
 
+# Invoke zoxide, safely escaping the arguments
+# PowerShell escapes command line arguments following the cmd.exe way instead of the C++/Rust way by default
+function zoxide {
+    $startInfo = [System.Diagnostics.ProcessStartInfo]::new("zoxide")
+    $startInfo.StandardOutputEncoding = [System.Text.Encoding]::UTF8
+    $startInfo.RedirectStandardOutput = $true
+    $startInfo.RedirectStandardError = $true
+    $startInfo.CreateNoWindow = $true
+    $startInfo.UseShellExecute = $false
+    if ($startInfo.ArgumentList.Add) {
+        # PowerShell 6+ uses .NET 5+ and supports the ArgumentList property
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Hey @davidkna,

Could you also provide an example of something that didn't work before this PR, but works now? Thanks!

davidkna:
Long paths:

```
❯ cd .\AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\
ResourceUnavailable:
Line |
  61 |      zoxide add $(__zoxide_pwd)
     |                   ~~~~~~~~~~~~
     | Program 'zoxide.exe' failed to run: StandardOutputEncoding is only supported when standard output is redirected.At line:61 char:5
+     zoxide add $(__zoxide_pwd)
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~.

```

Parsing errors (this one only fails in regular PowerShell for me and PowerShell Core works fine):
```
❯ mkdir 'slash at the end \'
❯ zoxide add 'slash at the end \'
zoxide: could not resolve path: C:\slash at the end "
```

PR #186, opened 2021-04-20
Use git2 to remove Git binary dependency

ajeetdsouza:


ajeetdsouza:
`git2` introduces a build dependency on `openssl.dev` and `pkg-config`, which I'd rather not have for something as simple as a version check. Closing.

PR #198, opened 2021-04-30
add armv7/aarch64 support to installer

adamcstephens:
This works on aarch64 for me, but I don't currently have an armv7 host to test.


ajeetdsouza:
Thanks!

PR #200, opened 2021-05-01
feature: cli completions

### LINKED ISSUE #9: Add zsh completion
It would be awesome to tab complete the top rated entries.

Side note: All the tickets are because your tool is awesome :) 🥇

XyLyXyRR:
Hi, there. I just impl completions for this project.

But there is a little awkward. clap 3-beta doesn't have 'completions' right now, so I downgrade to clap 2. (x
And a little ugly code to handle these differences.

Thank you.


ajeetdsouza:
Hey @XyLyXyRR, thanks for your work on this!

clap v3 does actually support shell completions. I'm not sure how good they are, but I suppose it wouldn't hurt to include a version-controlled directory with autogenerated completions from clap while we test them out. Unfortunately, that would require reworking your current implementation.

The way I would go about this is:

- Use the [`generate_to`](https://docs.rs/clap_generate/3.0.0-beta.2/clap_generate/fn.generate_to.html) function in `build.rs` to generate completions at compile time. The way it's done is admittedly a little hacky, but the exact procedure is given in the documentation.
- The `include!("src/cli.rs");` line requires the whole app to be specified in a single file with no crate dependencies.
  - Move all the current `App`-related structs out into `src/app.rs`.
  - Rename `crate::cmd::Cmd` to `crate::run::Run`.
  - Let `crate::run::<subcommand>` import from `crate::app::<subcommand>`.
- Use `clap_generate::generate_to` to write completions to `contrib/completions`. We can move it out of `contrib` once it's sufficiently stable. Use `include!(src/app.rs);` for getting the struct, and convert the struct to an app using [`into_app`](https://docs.rs/clap/3.0.0-beta.2/clap/trait.IntoApp.html#tymethod.into_app).

PS: the issue you linked actually refers to completions for `z` using `zoxide query -l`, which would not be fixed by this PR, so I've removed it from the PR description.

XyLyXyRR:
Hey @ajeetdsouza, I finished it.

Have a good day!

ajeetdsouza:
This is great, thank you!

PR #207, opened 2021-05-09
Prevent endless recusion with alias cd on fish

### LINKED ISSUE #145: Fish: Allow cd aliasing

dnlmlr:
This is a fix for the endless recursion on fish: fixes #145

- Create a custom __fish_cd function for use in __zoxide_cd.
- The __fish_cd function is set to be either a copy of the fish cd
  wrapper function, or an alias for `builtin cd` as fallback if the
  wrapper function does not exist for some reason.
- This prevents an endless recusion when using `alias cd=z` that would
  occur because __zoxide_cd would call the `cd` alias which would in
  turn call `z` and so on.

### DIFF
--- a/templates/fish.txt
+++ b/templates/fish.txt
@@ -5,6 +5,21 @@
 # Utility functions for zoxide.
 #
 
+# provide a custom __fish_cd function that is either a copy of the default fish 
+# cd wrapper function, or an alias for the `builtin cd` to prevent recursively 
+# calling cd when using `alias cd=z`.
+#
+# only create the function `__fish_cd` if it doesn't yet exist
+if ! functions -q __fish_cd
+    if functions -q cd
+        # use the fish wrapper function for __fish_cd if it exists
+        functions -c cd __fish_cd
+    else
+        # use `builtin cd` as fallback if the cd wrapper function does not exist
+        alias __fish_cd="builtin cd"
+    end
+end
+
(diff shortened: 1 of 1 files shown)


dnlmlr:
An endless recurive loop can still occur if `alias cd=z` is set __before__ invoking `zoxide init fish | source`, but other than that it should be working as intended. Since it didn't work at all before, I still think this is an improvement.

ajeetdsouza:
This is a great idea. Thank you!

dnlmlr:
Thanks for merging 👍 

There is one thing to consider. With your approach of unsetting the `__zoxide_cd_internal` instead of checking its existance, we get the recursion again when calling `zoxide init fish | source` after the `alias cd=z`. That situation might occur when accidently sourcing the `config.fish` more than once. 

Just something to keep in mind

PR #211, opened 2021-05-15
Support Fish's private mode

kidonng:
See fish docs: https://fishshell.com/docs/current/#private-mode

In this mode, fish will not record any command history. Some other projects also respect this config, for example [Fish port of z](https://github.com/jethrokuan/z/pull/98). I think it's a nice addition to zoxide!


ajeetdsouza:
Thanks! This is a great idea, I'll merge this once we get a reply on https://github.com/fish-shell/fish-shell/issues/8001.

PR #212, opened 2021-05-16
Improve Fish's completions

kidonng:
This PR brings two improvements to Fish's completions, when completing `{{cmd}}`:
- Let Fish only complete directories and excludes files
- If there is no match in the current directory, Fish will query zoxide with current token and serve them as completions:

  image

  This is approximately the same as using `zoxide query -i <current token>`, but more integrated with Fish.

Let me know if you have any suggestion over these two changes.

### DIFF
--- a/src/shell.rs
+++ b/src/shell.rs
@@ -1,5 +1,14 @@
 use crate::app::InitHook;
 
+const FZF_COMPLETE_OPTS: &str = "\
+    --bind=ctrl-z:ignore \
+    --exit-0 \
+    --height=35% \
+    --inline-info \
+    --no-sort \
+    --reverse \
+    --select-1";
+
 #[derive(Debug, Eq, PartialEq)]
 pub struct Opts<'a> {
     pub cmd: Option<&'a str>,
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -122,15 +122,7 @@ if [[ :"${SHELLOPTS}": =~ :(vi|emacs): ]] && [ "${TERM}" != 'dumb' ]; then
 
         if [[ ${query} == *"${trigger}" ]]; then
             query="${query:0:$(({{ "${#query} - ${#trigger}" }}))}"
-            COMPREPLY=("$(_ZO_FZF_OPTS="\
-                --bind=ctrl-z:ignore \
-                --exit-0 \
-                --height=35% \
-                --inline-info \
-                --no-sort \
-                --reverse \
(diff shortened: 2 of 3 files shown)


ajeetdsouza:
Thanks! Two small issues I had when trying this:

- If I type in `z foo bar` and hit Tab, I only get results for `bar` (`foo` is ignored).
- If I type in `z foo bar` and use the autocomplete, it will complete only `bar`, leaving me with `z foo /home/bar`. Since this contains multiple arguments, it doesn't cd correctly.

kidonng:
> - If I type in `z foo bar` and hit Tab, I only get results for `bar` (`foo` is ignored).

Only the token under the cursor is searched, see reasons below.

> - If I type in `z foo bar` and use the autocomplete, it will complete only `bar`, leaving me with `z foo /home/bar`. Since this contains multiple arguments, it doesn't cd correctly.

That's just hoe tab completion works. You can't expect it to change anything else than the current token. That's why it doesn't take any other token into consideration, it just won't make sense.

If someone do want to use multiple tokens, they should use `zoxide query -i` instead.

ajeetdsouza:
I understand that, but IMO completions should never result in a command that is invalid, which is what's happening here.

Btw, merging this PR might take a while - I want to try a couple of different solutions to see what works best.

kidonng:
> but IMO completions should never result in a command that is invalid

That means every completion should implement a parsing function to handle all the cases a program may encounter, it which is obviously not practical.

Technically we can prevent providing completion if user is completing the second argument, but IMO that's just unnecessary. You can easily reproduce the same issue with regular `cd`, and I don't anyone have an issue with it.

ajeetdsouza:
@kidonng if you have time, do take a look at https://github.com/ajeetdsouza/zoxide/pull/257. I've only implemented Bash completions at the moment, but I'd love to know what you think:

- Completions only work when there is one word (therefore completions are always valid).
- By default, it uses the same completions as `cd`.
- If the query ends with `**`, it opens up `fzf` and does an interactive search.

If it works well, we can hopefully write completions for other shells in line with this.

ajeetdsouza:
Hey @kidonng, I've updated the PR. This is what the new completion function does:

- It completes directories by default.
- If there is more than one argument, or if the last argument ends in `**`, it makes an interactive query via `fzf`.
- Early exit from interactive query removes the `**` if present.

I'd appreciate any thoughts on this.

PR #213, opened 2021-05-17
Remove deleted entries if not accessed in the last 90 days

ajeetdsouza:


kidonng:
I was going to comment before you merge, but something else distracted me 😅
My concern is that, the entry must be at a incredibly high score to not be cleaned by aging in 90 days. I assume most unused entries will be cleaned in a week, for a regular user. Does this added feature actually works?

ajeetdsouza:
I think the answer to this is - it depends on how you use zoxide.

- If you use a lot of temp directories while working, you might have a large number of entries in your database that no longer exist.
- If you don't want zoxide to forget, you might have set _ZO_MAXAGE to a very high value. The default value is also fairly high at `10000`, and some users may not reach it for a while.
- Deleted entries are rarely a problem for the `z` command, since it only needs to find the first match (which is usually not deleted). However, commands like `zi` or `zoxide query -l` would slow down with more directories, since they have to make a lot of `stat()` system calls.

On my system, the number of directories in my database dropped by half after merging this PR.

```shell
zoxide query --list --all | wc -l # original number of directories
# build zoxide on the latest master

zoxide query --list # this will trigger a scan of the entire database
zoxide query --list --all | wc -l # new number of directories
```

PR #222, opened 2021-06-05
chore: add ohmyzsh to readme

pleshevskiy:
Added information to Readme on how to install the zoxide plugin in the Oh My Zsh framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,22 @@ Add this to your configuration (usually `~/.zshrc`):
 eval "$(zoxide init zsh)"
 ```
 
+If you are using the [Oh My Zsh](https://github.com/ohmyzsh/ohmyzsh) framework,
+you can add the plugin with the following command
+
+```sh
+mkdir -p "$ZSH/plugins/zoxide" && zoxide init zsh > "$ZSH/plugins/zoxide/zoxide.plugin.zsh"
+```
+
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Hey @pleshevskiy! I'm a little confused - why do you want to create an OMZ plugin here?

The idea of zoxide being bundled with its own init script is that there is never a version mismatch between the zoxide binary and the zsh script. Moreover, you'd have to remember to recreate the script whenever you upgrade zoxide, which seems less than ideal.

pleshevskiy:
Thanks for the clarification! I prefer not to change the .zshrc but add plugins that extend the configs. I think the right thing to do would be to add `eval` to a plugin.

ajeetdsouza:
Sure, you can add a file containing `eval "$(zoxide init zsh)"` as an OMZ plugin. Even with this approach, you can't have a standard "zoxide plugin" file -- if you want to customize any flags when calling `zoxide init`, you'll have to change that file anyway. Performance-wise, you'll have to access an extra file on startup, but nothing too impactful.

I'll close this PR, since the usual approach seems to be more straightforward. However, do feel free to ping me on this thread if you have any more questions.

ajeetdsouza:
@pleshevskiy zoxide is now officially a part of `ohmyzsh`, you no longer need to write a custom plugin for this.

PR #230, opened 2021-07-06
Add --exit-0 to fzf

xeruf:
So there is no prompt if there is no match

### DIFF
--- a/src/fzf.rs
+++ b/src/fzf.rs
@@ -16,7 +16,7 @@ impl Fzf {
         if multiple {
             command.arg("--multi");
         }
-        command.arg("--exit-0").arg("-n2..").stdin(Stdio::piped()).stdout(Stdio::piped());
+        command.arg("--exit-0").arg("--preview='ls -al --color --human-readable --group-directories-first --file-type {2..}'").arg("-n2..").stdin(Stdio::piped()).stdout(Stdio::piped());
         if let Some(fzf_opts) = config::fzf_opts() {
             command.env("FZF_DEFAULT_OPTS", fzf_opts);
         }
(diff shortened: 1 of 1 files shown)


xeruf:
Second commit adds a preview

ajeetdsouza:
Hey @xeruf, you might want to use the `_ZO_FZF_OPTS` environment variable (see `zoxide --help` for details). Regarding the changes you've suggested, I don't think we can merge them because:

- The preview window looks great, but it doesn't work on Windows. However, I might add it to the wiki as a "common configuration".
- The `--exit-0` switch may not work for everyone - for example, I mainly use `zi` for browsing the database, and the experience can be jarring if I'm expecting an interactive window to open up but it doesn't.

I'll close this PR for now, but I'd love to hear any thoughts you have.

xeruf:
>     * The preview window looks great, but it doesn't work on Windows. However, I might add it to the wiki as a "common configuration".

I guess we could test whether `ls` is available and otherwise use whatever there is on windows instead.

>     * The `--exit-0` switch may not work for everyone - for example, I mainly use `zi` for browsing the database, and the experience can be jarring if I'm expecting an interactive window to open up but it doesn't.

so you are expecting an empty window to open?

ajeetdsouza:
> I guess we could test whether ls is available and otherwise use whatever there is on windows instead.

On my system, I use `exa` instead of `ls`. This is why I want to leave this as a user-configurable option, rather than hardcoding it ourselves.

> so you are expecting an empty window to open?

I suppose it's the equivalent of opening an empty file with `less`. For that matter, even `--select-1` might have been useful, but for someone who just wants to browse the database, it doesn't work as well.

---

For now, I think a better way would be to add these flags as a section under `Configuration` in the README, and let users customize `_ZO_FZF_OPTS` themselves.

PR #235, opened 2021-07-19
Add "matching" to README

lapo-luchini:
README was linking to #matching but was not defined.
Adapt content from man-page to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,21 @@ Be sure to set these before calling `zoxide init`.
   - When set to `1`, `z` will resolve symlinks before adding directories to the
     database.
 
+## Matching
+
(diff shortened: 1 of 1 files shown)


lapo-luchini:
Whops, I failed to notice that actual link is not to `#matching` but rather `wiki/Algorithm#matching`.
Problem is then: wiki is disabled on GitHub and this link is broken.

ajeetdsouza:
Hey @lapo-luchini,

I must have disabled the wiki by accident, I've enabled it again. I'll close this PR, since I'd like to keep the README as small as possible. I'll be moving more stuff out of the README and into the wiki soon.

Thanks for pointing this out!

PR #239, opened 2021-07-22
Fix: get PWD from xonsh shell env instead of the os env

### LINKED ISSUE #238: [xonsh] get $PWD from 'shell' instead of the 'os' environment
In a **xonsh** shell **zoxide** tries to get the current directory from the OS environment variable `PWD`, which isn't updated unless a special option is defined (and doesn't even exist in a login xonsh shell, which leads to `$PWD not found in env` errors) https://github.com/ajeetdsouza/zoxide/blob/d74cd962f05373d3d8a8b3959ea22db8f70c0452/templates/xonsh.txt#L40
As far as I understand it, a better approach would be to use the shell environment variable `$PWD` instead, which is updated on each directory change without the need to set any extra options.

#### Some background info:
In xonsh there are two environments:
1) the **shell** environment (`__xonsh__.env` or `${...}`; [see this tutorial](https://xon.sh/dev/tutorial.html?#the-environment-itself))
2) the **os** environment from Python (`os.environ` or via a wrapper [`xonsh.platform.os_environ`](https://github.com/xonsh/xonsh/blob/db99b646354460603e9c6d39563efd0e92ac565e/xonsh/platform.py#L455))

Unless a [special flag is manually set](https://xon.sh/dev/tutorial.html?#environment-variables), the OS environment isn't updated when the shell environment changes:
> Note To update os.environ when the xonsh environment changes set $UPDATE_OS_ENVIRON to True.

eugenesvk:
fixes #238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Nushell: errors on 0.33.0.
 - PowerShell: errors when initializing in `StrictMode`.
 - Bash/POSIX: remove conflicting alias definitions when initializing.
+- Bash: remove extra semicolon when setting `$PROMPT_COMMAND`.
+- Xonsh: use shell environment instead of `os.environ`.
 
 ## [0.7.2] - 2021-06-10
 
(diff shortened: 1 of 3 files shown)


eugenesvk:
Thanks for fixing the `builtins`, I was just looking at your template, but couldn't figure out which condition would fit for the `from builtins import __xonsh__` line, glad that you've just got rid of all of them :)

ajeetdsouza:
Haha, no problem! Thanks for finding and fixing this issue 😄

PR #240, opened 2021-07-25
Use single underscore for fish function/variables

kidonng:
Here is a little background (though it's 4 years ago): https://gitter.im/fish-shell/fish-shell?at=5730b1bcf16c08510662085b

This is an opinionated thing and it may break some advanced user's setup, but IMO it's worth it to be consistent with most Fish plugins out there.

Fish itself only uses two underscores for "private" functions (except for `_validate_int`, which is being tackled with fish-shell/fish-shell#8168). It's nice to be able to distinguish them with others.

### DIFF
--- a/templates/fish.txt
+++ b/templates/fish.txt
@@ -6,7 +6,7 @@
 #
 
 # pwd based on the value of _ZO_RESOLVE_SYMLINKS.
-function __zoxide_pwd
+function _zoxide_pwd
 {%- if resolve_symlinks %}
     builtin pwd -P
 {%- else %}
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Hey @kidonng, thanks for the PR!

According to what you linked, the `__zoxide_` prefix should only be used for naming private functions and variables, which is exactly what we're doing here -- we only really want to expose the `z` and `zi` functions. The `__zoxide_` functions are only defined that way for advanced users who'd like to tinker with things.

Even if that wasn't the case, I'd like to keep the naming conventions uniform across all the shells unless absolutely necessary, so I won't merge this. Sorry!

kidonng:
It is only about changing to use a single underscore to be consistent with most Fish plugins out there and have a different convention than Fish's own two underscores.

But not changing it is OK, won't really affect anything.

PR #289, opened 2021-10-28
Fallback to using CDPATH if there are no matches

phillipwood:
I've just started using zoxide and while I find it useful I'm missing being able to rely on CDPATH in fish for the paths that zoxide does not know about yet. I've only implemented this for fish so far as that's what I'm using but I can certainly add CDPATH support for bash and zsh if you like the idea (I'm not sure if the other shells that zoxide supports have CDPATH support)


ajeetdsouza:
I'm not sure about this one. The purpose of zoxide is to ensure that you don't need things like CDPATH. Granted that you have to cd into a new directory manually the first time, but you can always "jump" to it after that.

I'll close this for now, but I'm open to reconsidering it if there's enough demand for a feature like this.

Issue #299, opened 2021-11-10
Aliases not working in v0.7.9


Aankhen:
(PowerShell 7.2.0, Windows 10 x64. Also failed on the last two 7.1.x versions of PowerShell.)

With no change in my configuration, zoxide’s aliases are no longer defined after upgrading from v0.7.8 to v0.7.9. This is what I run in a module imported by my profile script:

```powershell
$script:zoxide_char = 'p'

Invoke-Expression (& {
    $hook = if ($PSVersionTable.PSVersion.Major -lt 6) { 'prompt' } else { 'pwd' }
    (zoxide init --hook $hook --cmd $script:zoxide_char powershell) -join "`n"
})
```

The relevant change in the output of <kbd>zoxide init --hook pwd --cmd p powershell</kbd> is that `-Option AllScope` was added to the aliases in ed47c2b1ac4420d8493b911b10b6e059f8af6691, so it produces this code:

```powershell
Set-Alias -Name p -Value __zoxide_z -Option AllScope
Set-Alias -Name pi -Value __zoxide_zi -Option AllScope
```

Looking at the docs, it makes sense that I wouldn’t see the aliases in my shell, since the expressions are being evaluated in a module imported by my profile script, i.e. in a child scope:

> Variables and aliases have an **Option** property that can take a value of **AllScope**. Items that have the **AllScope** property become part of any child scopes that you create, although they are not retroactively inherited by parent scopes.

Perhaps it would be better to pass ~~`-Option Global`~~ `-Scope Global` instead.

ajeetdsouza:
I'm on PowerShell 7.2.0, but I'm not seeing a `Global` option:

```
PS /Users/ajeet> Set-Alias -Name p -Value __zoxide_z -Option Global

Set-Alias: Cannot bind parameter 'Option'. Cannot convert value "Global" to type "System.Management.Automation.ScopedItemOptions". Error: "Unable to match the identifier name Global to a valid enumerator name. Specify one of the following enumerator names and try again:
None, ReadOnly, Constant, Private, AllScope, Unspecified"
```

ajeetdsouza:
Thanks @Aankhen, would you be able to take a look at #300 and see if it fixes the issue?

PR #300, opened 2021-11-10
PowerShell: use global scope for aliases

ajeetdsouza:
Fixes https://github.com/ajeetdsouza/zoxide/issues/299.


Aankhen:
Yup, that fixes it. Thanks very much for the prompt response!

PR #302, opened 2021-11-19
fix: don't error during __zoxide_unset

tekumara:
If `set -e` is enabled then `eval $(zoxide init zsh)` errors when the alias aren't already set (eg: in a fresh shell). This fixes that.


ajeetdsouza:
This should have been caught in CI, but there was a bug in the zsh test. Thanks for pointing this out!

PR #313, opened 2021-12-12
Add path escapes for powershell

### LINKED ISSUE #312: [Powershell] zoxide Does Not Escape Paths
## Problem
Powershell requires escaping special characters even if they're enclosed in single quotes. So calling `Set-Location` on, say `'D:\test\[foo] bar\'` will fail like:
```
Set-Location : Cannot find path 'test/[foo] bar' because it does not exist.
At line:1 char:1
+ Set-Location 'test/[foo] bar'
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (test/[foo] bar:String) [Set-Location], ItemNotFoundExcep
   tion
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.SetLocationCommand
```
image

## Solution
Path passed to Set-Location should be escaped:

image

ditsuke:
Fixes #312

Adds path escape to `Set-Location` call, fixing compatibility with paths containing special characters like square brackets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - PowerShell: use global scope for aliases.
 - Zsh: fix errors with `set -eu`.
 - Fzf: handle early selection.
+- PowerShell: correctly handle escape characters in paths.
 
 ## [0.7.9] - 2021-11-02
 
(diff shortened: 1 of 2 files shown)


ajeetdsouza:
Turns out `Set-Location` also takes in a `-LiteralPath`, so I used that instead. Thanks for your contribution!

ditsuke:
> Turns out `Set-Location` also takes in a `-LiteralPath`, so I used that instead. Thanks for your contribution!

I realized only after committing. Thanks for accepting!

PR #330, opened 2022-01-07
Document that webinstall.dev/zoxide installs more than zoxide

### LINKED ISSUE #318: Document that webinstall.dev/zoxide installs more than zoxide
Reviewing the script returned from `curl -sS https://webinstall.dev/zoxide` ([gist](https://gist.github.com/loganlinn/e2460478a77262e63a374384139dec38)), I was surprised to see that it installs additional tools (`webi`, `pathman`, `envman`, more?) to the environment that don't get cleaned up _and_ make unannounced changes to `~/.bashrc`, `~/.zshrc`, etc.

I understand these sorts of installers are not for everyone and the benefits the webinstall project provides to OSS maintainers, but I have a strong feeling this sort of low-visibility "tool bundling" is unexpected behavior to most.

dharmaturtle:
Resolves #318

I was not pleased when I blindly ran `webinstall`. Had I used more braincells I would've elected to use a package manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ To install zoxide, run this command in your terminal:
 curl -sS https://webinstall.dev/zoxide | bash
 ```
 
-Alternatively, you can use a package manager:
+The above installs `webi`, `pathman`, `envman`, and makes changes to your `~/.bashrc` and/or `~/.zshrc`. For a less intrusive option, use a package manager:
 
 | Distribution       | Repository              | Instructions                                                                                   |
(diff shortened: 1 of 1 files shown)


dharmaturtle:
Also, does it make sense to add

>`zoxide` works across all major platforms. If your distribution isn't included
in the list below, you can download the binary from the [Releases] page and
copy it to your `$PATH`.

back to the readme? `apt install zoxide` doesn't work for Ubuntu 20, which is the current LTS.

ajeetdsouza:
Hey, zoxide now ships its own installer once again. Sorry this took so long!

PR #334, opened 2022-01-12
Use elvish v0.17 lambda syntax in completions generation

### LINKED ISSUE #323: Elvish Shell Syntax Change
One of the shells zoxide supports, elvish (https://elv.sh), is planning to deprecate the lambda syntax used. The new syntax was supported starting in v0.17, but will be mandatory in v0.18.

v0.18 is planned for release [03/2022](https://github.com/elves/elvish/blob/master/0.18.0-release-notes.md). The output for `zoxide init elvish` needs to be updated before then. It is worth noting that the fix will break 0.16 support.

I'll submit a PR that has the breaking change. If it's an issue, maybe we could introduce something like `zoxide init --version 17 elvish` although that feels a little obtuse to me.

cherryblossom000:
Fixes #323

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Rename `_z` completion function to avoid conflicts with other shell plugins.
+- Elvish: upgrade to new lambda syntax.
 
 ## [0.8.0] - 2021-12-25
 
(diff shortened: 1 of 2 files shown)


ajeetdsouza:
Thank you for your contribution!

cherryblossom000:
I think the tests are failing because you need to update nixpkgs (the version that this project currently uses has Elvish v0.16.3).

ajeetdsouza:
Yeap, I'm doing just that.

PR #342, opened 2022-01-24
--interactive/-i and --score/-s do not conflict in Query::query()

c02y:


ajeetdsouza:
I think this idea should be extended further -- `-i` should not show scores unless `-s` is present.

PR #346, opened 2022-02-01
Improve fzf readability on narrow screens

aisamu:
[--keep-right] trims paths from the left/start, preventing long folder
chains from taking over the prompt and making the choice ambiguous:

Without it:
```
/long/folder/chain/with/same/prefi...
/long/folder/chain/with/same/prefi...
/long/folder/chain/with/same/prefi...
/long/folder/chain/with/same/prefi...
```

With it:
```
...der/chain/with/same/prefix/folder1
...der/chain/with/same/prefix/folder2
...der/chain/with/same/prefix/folder3
...der/chain/with/same/prefix/folder4
```

Thanks for taking the time to write this awesome utility!

[--keep-right]: https://www.mankier.com/1/fzf#--keep-right

### DIFF
--- a/src/fzf.rs
+++ b/src/fzf.rs
@@ -26,7 +26,7 @@ impl Fzf {
                 "--bind=ctrl-z:ignore",
                 "--exit-0",
                 "--height=40%",
-                "--inline-info",
+                "--info=inline",
                 "--no-sort",
                 "--reverse",
                 "--select-1",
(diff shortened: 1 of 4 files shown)


src/fzf.rs
```diff
@@ -25,11 +25,12 @@ impl Fzf {
             command.args(&[
                 "--bind=ctrl-z:ignore",
                 "--exit-0",
+                "--select-1",
                 "--height=40%",
-                "--inline-info",
+                "--info=inline",
                 "--no-sort",
                 "--reverse",
-                "--select-1",
```
ajeetdsouza, line 32: Thanks! Why was this line removed?

aisamu:
It shouldn't have been! Just moved upwards

PR #370, opened 2022-03-25
Simplify fish init script

kidonng:
- Remove usage of `__zoxide_unset`.
  - There is no need to remove namesake variables, which don't even make sense.
  - Similarly, no need to remove abbrs. Users should know better than to set an abbreviation with the same name.
  - New functions will discard old ones with the same name.

- Replace `alias` with idiomatic `function`.

  fish's [`alias`](https://fishshell.com/docs/current/cmds/alias.html) is simply a function wrapper and slower than defining functions directly.

### DIFF
--- a/templates/fish.txt
+++ b/templates/fish.txt
@@ -102,20 +102,15 @@ end
 {%- match cmd %}
 {%- when Some with (cmd) %}
 
-# Remove definitions.
-function __zoxide_unset
-    set --erase $argv >/dev/null 2>&1
-    abbr --erase $argv >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
> Users should know better than to set an abbreviation with the same name.

Sometimes these abbreviations are created by plugins, and are unknown to the user. It's happened multiple times in the past, so zoxide tries to cleanly override any previous definitions. I'm only using `__zoxide_unset` for commands like `z` and `zi`, I do not expect a user to have another definition of `__zoxide_z` for obvious reasons.

PR #371, opened 2022-03-30
switch elvish template to new try/catch syntax

skwerlman:
'except' has been deprecated and replaced by 'catch' in elvish 0.18, and will be removed in 0.19

see the relevant issue here: https://github.com/elves/elvish/issues/1497


ajeetdsouza:
Thank you!

PR #382, opened 2022-04-25
Limit nix features

rtzoeller:
This removes memoffset as an indirect dependency, and should slightly decrease build times.


ajeetdsouza:
Thanks!

PR #385, opened 2022-04-27
fix: interactive completion for fish

### LINKED ISSUE #341: Can't select in interactive completion on fish shell
when type `z foo<space><tab>` fzf opens and after selected one, it doesn't show on commandline

it works on zsh

NextAlone:
Just as mentioned in #341, use `\#` to indicate full path for zoxide to query.

In my tests it worked well.

resolved #341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fish: paths on Cygwin.
+- Bash/Fish/Posix/Zsh: paths on Cygwin.
 - Fish: completions not working on certain systems.
 
 ## [0.8.1] - 2021-04-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fish: fix paths on Cygwin.
+- Fish: paths on Cygwin.
+- Fish: completions not working on certain systems.
 
 ## [0.8.1] - 2021-04-23
 
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -13,7 +13,7 @@ function __zoxide_pwd() {
     \builtin pwd -P
 {%- else %}
     \builtin pwd -L
-{%- endif %} || \builtin true
+{%- endif %}
 }
 
 # cd + custom logic based on the value of _ZO_ECHO.
@@ -34,6 +34,7 @@ function __zoxide_cd() {
 {%- if hook == InitHook::Prompt %}
 function __zoxide_hook() {
     \builtin local -r retval="$?"
+    # shellcheck disable=SC2312
     \command zoxide add -- "$(__zoxide_pwd)"
     return "${retval}"
 }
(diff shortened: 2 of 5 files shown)


NextAlone:
Thanks for your work.

hferreira23:
@ajeetdsouza when do you expect to release a new version with this fix (pretty please)?

antoyo:
> @ajeetdsouza when do you expect to release a new version with this fix (pretty please)?

I'm wondering the same. It would be nice to have zoxide works correctly with fish.
Thanks for your work.

NextAlone:
> > @ajeetdsouza when do you expect to release a new version with this fix (pretty please)?
> 
> I'm wondering the same. It would be nice to have zoxide works correctly with fish.
> Thanks for your work.

maybe you install an git version for now？

ajeetdsouza:
@hferreira23 @antoyo I just released v0.8.2.

Zesko:
The version 0.8.2 works now for me using Fish shell version 3.4.1. Thanks!

I don't know if it works with Fish shell 3.5.0.

ajeetdsouza:
Yeap, seems to be working fine for me.

antoyo:
I just tested version 0.8.2 and it seems to misbehave.
For instance, when I type `z tmp <TAB>` it completes to: `z tmp z\#/tmp`.
Do I need to do anything after an upgrade to make it work?
I use fish version 3.4.1 with zoxide v0.8.2.

NextAlone:
> I just tested version 0.8.2 and it seems to misbehave.
> For instance, when I type `z tmp <TAB>` it completes to: `z tmp z\#/tmp`.
> Do I need to do anything after an upgrade to make it work?
> I use fish version 3.4.1 with zoxide v0.8.2.

it means only this path match

antoyo:
> it means only this path match

So, the behavior changed?
It used to complete to `z /tmp`; why is there the `z\#` prefix and why doesn't it delete the query (`tmp`)?

NextAlone:
> > it means only this path match
> 
> So, the behavior changed? It used to complete to `z /tmp`; why is there the `z\#` prefix and why doesn't it delete the query (`tmp`)?

Oh, yes, the method before doesnt work on some system

antoyo:
So, you're saying this behavior is expected?

Any way to at least delete the query `tmp` so that pressing enter actually go to the directory that was completed?

Zesko:
`z#/tmp` without `\` like the description in ReadMe.
Screenshot_20220627_145741

VS.

`z\#/tmp` with `\`
Screenshot_20220627_145712

`\#` seems the color.  I didn't change the color at all. 
But I think Fish may recognized `\#` and defined color for it.

NextAlone:
because fish will escape `\#` to `#`, and single `#` is not allowed

antoyo:
> Pressing enter will go to the directory that was completed. The `z#` prefix ensures that this happens.

Unfortunately, it doesn't: I get the following error:
```
~ $ z tmp z\#/tmp
zoxide: no match found
```
If I delete the query, it does work:
```
~ $ z z\#/tmp
/tmp $
```
But I'd rather avoid delete the query manually.

NextAlone:
➜ fish --version
fish, version 3.5.0-47-gfee5a9125

➜ zoxide --version
zoxide v0.8.2

Zesko:
Screenshot_20220627_151127

It works for me using Fish shell 3.4.1 with Starship too.

antoyo:
Oh, the issue comes from my alias of `z`. It indeed works if I remove that alias.

ajeetdsouza:
The `z\#` is now a `z!`, which looks a lot cleaner. Unfortunately, I don't think Fish supports a way to quote the completion rather than escape it, so instead of `z "z!/some directory/"` we end up getting `z z!/some\ directory/`.

PR #419, opened 2022-07-08
Add PowerShell ArgumentCompleter

### LINKED ISSUE #344: Tab completions for PowerShell
### Discussed in https://github.com/ajeetdsouza/zoxide/discussions/343

<div type='discussions-op-text'>

<sup>Originally posted by **@charlesroper** January 31, 2022</sup>
I've just installed Zoxide for PowerShell (7.2.1) and just wondering: are there no tab completions? Thanks!</div>

twobiers:
Resolves #344 

I have tried to implement an equivalent Tab completion for PowerShell, which was to some point successful. 
It works like the following:
- Use the default directory completion when no arguments are provided
- If an argument is provided, use it as zoxide query input

Unfortunately it doesn't seem to be possible to run interactive queries in PowerShell within the argument completer, therefore I don't know whether it is possible to implement the "space+tab" completion. 
Also I am unsure what should happen if mutliple arguments are provided. At the moment the tab completion ignores that and happily completes the second argument. For example `z dir1 dir2` would try to complete `dir2` to `z dir1 P:/tmp/dir2`.

Maybe it would be useful to also add the default completions into the completions, so that I could use `z` as a `cd` replacement.

Please let me know your thoughts.

This GIF shows, how it works:
WindowsTerminal_caWG2ym896

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -113,6 +113,52 @@ function __zoxide_zi {
 Set-Alias -Name {{cmd}} -Value __zoxide_z -Option AllScope -Scope Global -Force
 Set-Alias -Name {{cmd}}i -Value __zoxide_zi -Option AllScope -Scope Global -Force
 
+filter __zoxide_escapeStringWithSpecialChars {
+    $_ -replace '\s|#|@|\$|;|,|''|\{|\}|\(|\)|"|`|\||<|>|&','`$&'
+}
+
+Register-ArgumentCompleter -Native -CommandName "{{cmd}}" -ScriptBlock {
+    param(
+        $WordToComplete,
+        $CommandAst,
+        $CursorPosition
+    )
+
+    # Get the current command line and convert into a string
+    $Command = $CommandAst.CommandElements
+    $Command = "$Command"
+
+    # The user could have moved the cursor backwards on the command-line.
+    # We only show completions when the cursor is at the end of the line
+    if ($Command.Length -gt $CursorPosition) {
+        return
+    }
+
+    $Program,$Arguments = $Command.Split(" ",2)
+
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Hey, looks like this PR got closed by accident because we moved from `next` back to `main` (it was causing quite a bit of confusion, we started seeing PRs to both `next` and `main`). Sorry about that!

I tried out a rudimentary completer, this seems to work fine on pwsh 7.5.1 / macOS (although I don't know if it would work on Windows):

```powershell
Register-ArgumentCompleter -Native -CommandName z -ScriptBlock {
    param(
        $WordToComplete,
        $CommandAst,
        $CursorPosition
    )
    $result = zoxide query -i
    Set-Location $result

    # Clear the current line
    [Microsoft.PowerShell.PSConsoleReadLine]::RevertLine()
    [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
}
```

The way completions are done in other shells is:

- if the user is completing their first arg (or if there are no args), we use regular cd completions
- if the user presses Space+Tab (that means there are one or more args, and they have started a fresh arg), we use interactive completions

Would you be able to create a fresh PR that matches the behavior of existing shells?

twobiers:
Hi @ajeetdsouza Thanks for coming back to this. I'm not actively using windows anymore either, but can still try to test it again. Your proposed behavior is actually the same implemented (See the GIF). If you can change the target of this PR to main and reopen the PR, we can just reuse this one. If that's not possible, I can also create a fresh one.

ajeetdsouza:
Sadly, I'm not able to change the target either.

twobiers:
I think you would have to reopen the pull request to do so as targets can only changed on opened PRs.

Issue #392, opened 2022-05-17
Feature Request: add support for increment/decrement subcommands


JS-Zheng:
# Background
Our team has a large project, which has multiple branches located in different paths (has the same directory structures), for example: 
- /branch_a/device/driver/
- /branch_b/device/driver/
- /branch_c/device/driver/

Therefore, a directory name is not an effective factor for this case, adjusting rank/weight is a very useful feature for us.

# FR
Add an **inc(rement)** subcommand and a **dec(rement)** to update rank/weight manually.

For example:
```sh
# increment rank by default value
zoxide inc /branch_a/device/driver/  

# decrement rank by default value
zoxide dec /branch_b/device/driver/     

# increment rank by the specified value
zoxide inc=10 /branch_c/device/driver/ 
```

ajeetdsouza:
Have you tried the `<SPACE><TAB>` completions that ship with zoxide v0.8.0+? They seem ideal for this kind of structure.

In case you're not familiar with how it works, there's a GIF in the [README](https://github.com/ajeetdsouza/zoxide#getting-started).

JS-Zheng:
Yes, I usually use `<SPACE><TAB>` fzf completions. However,  I want to increase the rank of the recently developed branches, or reduce the rank of the outdated branches that has entered the maintenance cycle, so that the order of the fzf completions is more effective.

ajeetdsouza:
Closing in favour of https://github.com/ajeetdsouza/zoxide/issues/453.

Issue #236, opened 2021-07-20
Update database


AckslD:
If I renamed some of my common directories, is it possible to edit the database in some way to reflect this without loosing the current scores?

AckslD:
Thanks for your reply @ajeetdsouza! That would indeed be great I think!

kidonng:
@ajeetdsouza Would you reconsider #133? Once it's implemented, updating database can be archived via export -> edit -> import, and that can be implemented via simple shell scripts.

Allaman:
For anyone finding this issue with a similar problem (migrating from Linux to MacOS and therefore different paths) my workaround was as follows:
1. Export all directories via `zoxide query --all` on my old machine 
2. Running a search and replace over the output.
3. Add the paths on my new machine with the following script. Of course I will loose my scores but I can live with that :) 

```sh
#!/bin/bash

input="$HOME/Downloads/zoxide.txt"
if [[ -f "$_ZO_DATA_DIR/db.zo" ]]
then
  echo "WARNING zoxide db exists"
  echo "Continue?"
  select choice in "Yes" "No"; do
    case $choice in
      Yes ) echo "Going on"; break;;
      No ) exit;;
    esac
  done
fi

while IFS= read -r line
do
  if [[ ! -d "$line" ]]
  then
    echo "WARNING $line does not exist - not adding to database"
  else
    zoxide add "$line"
  fi
done < "$input"
```

ajeetdsouza:
Closing in favour of https://github.com/ajeetdsouza/zoxide/issues/453.

PR #422, opened 2022-07-11
Add manual score adjustment

TheRealFalcon:
This commit adds '-i' and '-d' options to the 'zoxide add' command.
'-i' increments the base score by a specified amount.
'-d' decrements the base score by a specified amount.

Also added unit tests covering the new functionality combined with
recency scoring.

Fixes/Supports #392

### DIFF
--- a/contrib/completions/_zoxide
+++ b/contrib/completions/_zoxide
@@ -30,6 +30,10 @@ _zoxide() {
         case $line[1] in
             (add)
 _arguments "${_arguments_options[@]}" \
+'(-d --decrement)-i+[Increment path(s) score by specified amount]:INCREMENT: ' \
+'(-d --decrement)--increment=[Increment path(s) score by specified amount]:INCREMENT: ' \
+'(-i --increment)-d+[Decrement path(s) score by specified amount. Score won'\''t go below 0]:DECREMENT: ' \
+'(-i --increment)--decrement=[Decrement path(s) score by specified amount. Score won'\''t go below 0]:DECREMENT: ' \
 '-h[Print help information]' \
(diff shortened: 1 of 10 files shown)


ajeetdsouza:
Hey @TheRealFalcon, sorry for the delayed reply. I think a feature like this would be very confusing to use.

The working of zoxide's algorithm is explained [here](https://github.com/ajeetdsouza/zoxide/wiki/Algorithm). I see you're setting the `last_accessed` field to `now`, which means that decrementing the rank can actually increase the score. Consider this scenario:

- There's a directory with `{ rank: 100, last_accessed: <last_year> }`. The effective score becomes `rank/4 = 25`.
- When you _decrement_ by `1`, the directory becomes `{ rank: 99, last_accessed: <few_seconds_ago> }`. Now, the score becomes `rank*4 = 396` (!!!)

Even if you changed the function to not edit the `last_accessed` field, it still means that changing the rank by `1` would change the effective score by anywhere between `0.25` and `4`.

I'm closing this PR for this reason, do feel free to comment with your thoughts.

TheRealFalcon:
Are you uninterested in supporting #392 at all, or is the issue just this implementation? If it's the implementation, I could try to change it so it works as you'd like, but if you're against the idea of #392 in general, it'd probably be best to close that issue.

TheRealFalcon:
@ajeetdsouza , thanks for the info. That's certainly something I could work on. I'm assuming #236 would be the best place to discuss details?

ajeetdsouza:
I opened up https://github.com/ajeetdsouza/zoxide/issues/453 with a spec, feel free to discuss there.

PR #442, opened 2022-08-16
docs: outdated guide in Nushell integration

azzamsa:
The command is not in sync with https://github.com/ajeetdsouza/zoxide/blob/b38b89eb3bd4f93f809665eca1d88aabc1ff1b34/templates/nushell.txt#L80 and https://github.com/ajeetdsouza/zoxide/blob/b38b89eb3bd4f93f809665eca1d88aabc1ff1b34/man/man1/zoxide-init.1#L38

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ Note: zoxide only supports fish v3.4.0 and above.
 Add this to your env file (find it by running `$nu.env-path` in Nushell):
 
 ```sh
-zoxide init nushell | save ~/.zoxide.nu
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
What you're seeing are the instructions for the next version of zoxide, which has not been released yet. To avoid this confusion in future, I've moved development to a `next` branch, so that the `main` branch always points to the latest release.

azzamsa:
Thanks for the idea. I will apply the same method for my repos. I have seen the similar confusion over and over again. Now it happened to me :)

Issue #338, opened 2022-01-20
Can't select interactive completions


CharlesARoy:
Hopefully I didn't miss a previous issue for this.

After installing zoxide with conda (and adding eval "$(zoxide init zsh)" to the bottom of my .zshrc file) I can't seem to select paths from the interactive completions. For example, after launching the interactive completions, I see:
![image](https://user-images.githubusercontent.com/46253001/150275014-3497dfba-6ffd-4bf2-9b69-6be139d5b99f.png)

Then, if I hit enter on the second option, I get this:
![image](https://user-images.githubusercontent.com/46253001/150275088-e022646d-1f76-4625-ab3e-b09d5c2de6c3.png)

Not sure what I'm doing wrong... Please let me know if you need any specific system information.

ajeetdsouza:
Are you able to select things with fzf?

CharlesARoy:
I'm not very familiar with fzf but if I run `fzf -m`, I can then scroll through the files that are returned and select any of them.

![image](https://user-images.githubusercontent.com/46253001/150389574-c91168b4-aa16-45b5-9244-1c6d4dc63c0a.png)

CharlesARoy:
Hmm, after removing zsh-autocomplete, I was able to select stuff normally with zoxide. These were my zsh-autocomplete settings prior to removing them (with the `eval "$(zoxide init zsh)"` command following this in my .zshrc file). If you use zsh-autocomplete, let me know if you spot any obvious issues with this setup. 

```
# The code below sets all of `zsh-autocomplete`'s settings to their default
# values. To change a setting, copy it into your `.zshrc` file.

zstyle ':autocomplete:*' default-context ''
# '': Start each new command line with normal autocompletion.
# history-incremental-search-backward: Start in live history search mode.

zstyle ':autocomplete:*' min-delay 0.05  # float
# Wait this many seconds for typing to stop, before showing completions.

zstyle ':autocomplete:*' min-input 0  # int
# Wait until this many characters have been typed, before showing completions.

zstyle ':autocomplete:*' ignored-input '' # extended glob pattern
# '':     Always show completions.
# '..##': Don't show completions when the input consists of two or more dots.

zstyle ':autocomplete:*' list-lines 16  # int
# If there are fewer than this many lines below the prompt, move the prompt up
# to make room for showing this many lines of completions (approximately).

zstyle ':autocomplete:history-search:*' list-lines 16  # int
# Show this many history lines when pressing ↑.

[...]

PR #449, opened 2022-09-03
Fix interactive completion with zsh-autocomplete

aaronkollasch:
## Problem
With [zsh-autocomplete](https://github.com/marlonrichert/zsh-autocomplete) and fzf installed, `z foo<SPACE><TAB>` opens the fzf picker as expected, but upon pressing `<ENTER>`, the prompt disappears and the fzf picker goes blank, with the text `0n` entered (screenshot below).

Last login Sat Sep 3 003804 on ttys008

Pressing escape three times returns me to the prompt, with no completion entered.

Pressing enter causes the error `zoxide: no match found` to be printed and fzf to reopen, again with the text `0n`.

## Cause
This is because zsh-autocomplete patches the `_complete` function (and others) to return an error if no new matches are found:
```zsh
% which _complete
_complete () {
	local -i nmatches=$compstate[nmatches]
	.autocomplete._complete "$@" || _autocomplete.ancestor_dirs "$@" || _autocomplete.recent_paths "$@"
	(( compstate[nmatches] > nmatches ))
}
```
and `__zoxide_z_complete` can be called by different completion functions after each one returns an error. Each time, this launches a new fzf picker before `\e[0n` can be input to the shell, so fzf catches it instead.

## Solution
If `__zoxide_z_complete` sees that zsh-autocomplete is present (e.g. by seeing that the patched function `.autocomplete._complete` is present), it should add an empty completion match. 

zsh-autocomplete will use the empty match and not run `__zoxide_z_complete` a second time, so `__zoxide_z_complete_helper` will work as expected.

## Alternative
[aaronkollasch:zsh-autocomplete-compat-2](https://github.com/aaronkollasch/zoxide/tree/zsh-autocomplete-compat-2)

Instead of checking for zsh-autocomplete and creating an empty completion, a more general solution is to send the zoxide result through zsh's completion system, like so:
```zsh
\builtin local result="${__zoxide_z_prefix}${__zoxide_result}"
compadd -Q "${(q-)result}"
```
Then, `__zoxide_z_complete_helper` only needs to reset the prompt.
I've tested this and it seems to work, both with zsh-autocomplete and using compinit with its default settings.

---

Fixes https://github.com/ajeetdsouza/zoxide/issues/338.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- zsh: completions clashing with `zsh-autocomplete`.
+
 ## [0.8.3] - 2022-09-02
 
 ### Added
--- a/templates/elvish.txt
+++ b/templates/elvish.txt
@@ -87,8 +87,10 @@ edit:add-var {{cmd}}~ $__zoxide_z~
 edit:add-var {{cmd}}i~ $__zoxide_zi~
 
 # Load completions.
-{# zoxide-based completions are currently not possible, because Elvish only
- # prints a completion if the current token is a prefix of it. -#}
+{#-
+  zoxide-based completions are currently not possible, because Elvish only prints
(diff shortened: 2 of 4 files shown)


ajeetdsouza:
@aaronkollasch I used your second implementation since it's a generalized approach -- could you verify that it works for you?

aaronkollasch:
Great, that works for me!

Issue #454, opened 2022-09-07
New preview window: ls illegal option


kaihowl:
Mac OS 10.15.7
Using zsh.
Zoxide 0.8.3

With "z" tab completion, I see the following in the preview window:

![image](https://user-images.githubusercontent.com/474193/188835526-a603a873-432e-477b-a940-6b2e833c7f8e.png)

I understand that this might be too little information to go on. Please let me know, what I can provide to help analyze the issue.

ajeetdsouza:
That's very strange, it works on macOS for me. What's the value of `which ls`?

kaihowl:
```
$ which ls
> /bin/ls
```

ajeetdsouza:
Fixed in https://github.com/ajeetdsouza/zoxide/pull/460.

kaihowl:
@ajeetdsouza thank you for fixing this! What are your plans for a new release? :)

aldanor:
Looking forward to the version bump as well as the `ls` error is pretty frustrating.

PR #460, opened 2022-09-16
Fix illegal option in fzf preview window

boomker:
#454


ajeetdsouza:
Only Linux allows `--color=always`, it's not a POSIX requirement and BSD operating systems don't include it. Strangely, it works for me under macOS (not sure why), but if it's causing problems on other systems we can remove it.

PR #461, opened 2022-09-22
Feature: zoxide edit subcommand

TheRealFalcon:
@ajeetdsouza , I have an implementation that works for some basic cases I manually checked. It still needs tests and polish, but I wanted to make sure that this approach looks generally ok before moving on.

What do you think?


src/db/mod.rs
```diff
@@ -32,15 +32,19 @@ impl<'file> Database<'file> {
 
     /// Adds a new directory or increments its rank. Also updates its last accessed time.
     pub fn add<S: AsRef<str>>(&mut self, path: S, now: Epoch) {
+        self.add_raw(path, now, 1.0)
+    }
+
+    pub fn add_raw<S: AsRef<str>>(&mut self, path: S, last_accessed: Epoch, rank: Rank) {
```
ajeetdsouza, line 38: No need of the original function, just add this parameter to this original `add` function.

src/cmd/edit.rs
```diff
@@ -0,0 +1,181 @@
...
+    Retry,
+    Exit,
+}
+
+impl Run for Edit {
+    fn run(&self) -> Result<()> {
+        let temp_dir = tempdir()?;
+        let temp_dir_path = temp_dir.path();
+        while let Some(db_edits) = get_db_edits()? {
+            let mut db_file = DatabaseFile::new(temp_dir_path);
```
ajeetdsouza, line 29: There's no need of using a tempfile here, `db.save()` already creates a tempfile and does an atomic file rename for you. Instead, the flow could be more like:
- Open the DB.
  ```rust
  let data_dir = config::data_dir()?;
  let mut db = DatabaseFile::new(data_dir);
  let mut db = db.open()?;
  ```
- Convert it to human-readable text, allow the user to edit it, and perform any validations.
- Clear the DB and write the new directories to it.
- `db.save()?`

ajeetdsouza (commented):
The approach looks good. I'd recommend using the `edit` crate over `dialoguer::Editor`, since it has support for a lot more editors across various platforms.

TheRealFalcon:
@ajeetdsouza , I addressed your comments and added some tests. I created a few additional commits for ease of reviewing assuming we can squash merge when things are ready.

I added some line validation tests, but testing the command end to end is tricky due to the terminal editor and the dialogue when things go wrong. I don't know of a way to test that without use of a mock library or some ugly dependency injection. If you'd like to pull in a mocking library, I don't mind adding a few more tests. Let me know what you'd like to do here.

src/cmd/edit.rs
```diff
@@ -0,0 +1,267 @@
...
+                None
+            }
+        };
+
+        let path: Option<String> = match path_txt {
+            Some(value) => {
+                if value.trim() != value {
+                    warnings.push(Problem::new(line_number, "path contains trailing whitespace".to_string()));
+                }
+                match resolve_path(&PathBuf::from(value)) {
```
ajeetdsouza, line 136: Let's not use `resolve_path` here. The user is directly editing the database, and is therefore expected to put in valid paths. If they're not valid, they'll just never show up in the results anyway. I'd say just test with `Path::is_absolute`.

src/cmd/edit.rs
```diff
@@ -0,0 +1,267 @@
...
+        println!("Zoxide database not altered");
+        Ok(())
+    }
+}
+
+fn get_db_edits() -> Result<Option<String>> {
+    let data_dir = config::data_dir()?;
+    let mut db = DatabaseFile::new(data_dir);
+    let mut db = db.open()?;
+    let mut stream = db.stream(util::current_time().unwrap());
```
ajeetdsouza, line 66: Streams run a bunch of filtering on the DB which we don't want to do here (we want to show everything in the DB, valid or not). I'd suggest just accessing the `dirs` field directly.

src/cmd/edit.rs
```diff
@@ -0,0 +1,267 @@
...
+                }
+            }
+            None => {
+                errors.push(Problem::new(line_number, "cannot parse 'path' field".to_string()));
+                None
+            }
+        };
+
+        if let (Some(path), Some(last_accessed), Some(rank)) = (path, last_accessed, rank) {
+            db.add(&path, last_accessed, rank);
```
ajeetdsouza, line 156: This just struck me, but we don't need `add` at all for this feature. The user is editing the entire database, so when they save their edits, we can just parse all the edits into a new `Vec<Dir>` and replace the old `db.dirs` with this one.

src/cmd/edit.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+fn get_db_edits() -> Result<Option<String>> {
+    let data_dir = config::data_dir()?;
+    let mut db = DatabaseFile::new(data_dir);
+    let mut db = db.open()?;
+    let mut stream = db.stream(util::current_time().unwrap());
+    let mut to_edit = String::from(HEADER);
+    while let Some(dir) = stream.next() {
+        writeln!(&mut to_edit, "{},{},{}", dir.last_accessed, dir.rank, dir.path)?;
```
ajeetdsouza, line 69: Rank is an `f64`. That's potentially a lot of decimals that the user doesn't really care about. We could just print the first 2 decimal places here and it should be fine.

src/cmd/edit.rs
```diff
@@ -0,0 +1,267 @@
...
+        let line_number = index + 1;
+        let first_char = line.trim().chars().next();
+        if let Some(first_char) = first_char {
+            if first_char == '#' {
+                continue;
+            }
+        } else {
+            continue;
+        }
+        let mut split = line.split(',');
```
ajeetdsouza, line 96: Paths might have commas in them. Use `splitn` to create 3 splits.

src/cmd/edit.rs
```diff
@@ -0,0 +1,267 @@
+use crate::cmd::{Edit, Run};
+use crate::db::{Database, DatabaseFile, Epoch, Rank};
+use crate::util::resolve_path;
+use crate::{config, util};
+use anyhow::Result;
+use std::fmt::Write as FmtWrite;
```
ajeetdsouza, line 6: If you don't need an import, you can import it `as _`.

ajeetdsouza (commented):
Hey, the tests look great. Left you some more feedback.

src/cmd/edit.rs
TheRealFalcon: This one seems a bit odd to me. We're working around the API to skip a check that might add a few milliseconds to processing thousands of directories.

TheRealFalcon:
Thanks @ajeetdsouza . I addressed all of your comments, and also commented inline about one, but I still applied it.

ajeetdsouza:
Hey @TheRealFalcon, sorry about the late response. I've been very tied up with work of late.

I thought of a new way to do an edit command, and I think it's a lot easier to use than our current editor-based approach. I've done a rough implementation at https://github.com/ajeetdsouza/zoxide/pull/498, I'd love to hear your thoughts on it!

Issue #459, opened 2022-09-15
[Feature Request] Jumping to children of current working directory


imlzo:
Feature request to add additional shell commands for jumping to children for working directory, similar to `jc` in [autojump](https://github.com/wting/autojump). 

WDYT about bringing this to zoxide's shell templates?

Happy to take this task on.

ajeetdsouza:
Sure! I'd suggest a feature that works this way:

- If the first parameter of `zoxide query` is `~`, then all results should start with `$HOME`.
- If the first parameter of `zoxide query` is `.`, then all results should start with `$PWD`.

You'd be able to use `z . foo` and it would only match a subdirectory.

Blonteractor:
Got a working backend for this. Right now I added a `zw` command which cds into a subdirectory by passing the `--workingdir` flag to `zoxide query`.

Blonteractor:
I got it working by making it a different command in the shell scripts, `zw`, don't know enough bash to implement checking the first parameter to `zoxide query` and triggering the respective behavior but it should be trivial, sound good?

ajeetdsouza:
This behaviour should be triggered within Rust rather than within Bash, since it would be common to all 8 shells supported by zoxide.

Blonteractor:
Aight made a [pull request](https://github.com/ajeetdsouza/zoxide/pull/468)

mtalexan:
I'd argue this should also be the default behavior for the `z` and `zi` commands, and a different command should be used for cases where the current directory context isn't relevant.

The `z.lua` tool (and for some cases `z` and `autojump`) seems to have understood that the following is the order for how common actions are:
1. Going to sub folder of the current
2. Going to the functional root (e.g. git root, hg root, etc) above the current
3. Going to a parent directory by name match
4. Going to an equivalent directory to the current, where some portion of the parent path is swapped for something else
5. (not handled by them) A recent folder not sharing most of the parent path

Option 5 already has a poor manual jump list in Linux with `pushd`, `popd`, `dirs`, `~1/`, `~2/`, etc.  

Currently I would estimate 98% of use cases a jump command wants to do something relative to their current folder (up, down, or over).  Only a small number of cases want to jump to unrelated folders.  Unfortunately, Zoxide currently _only_ supports jumping to unrelated folders, and custom functions would need to be written to enable jumping to folders relative to the current (e.g. a custom function that automatically inserts the current path as the first search term).

PR #468, opened 2022-10-17
Option to query only into sub directories in zoxide query

Blonteractor:
Adds feature(s) requested in [Issue#459](https://github.com/ajeetdsouza/zoxide/issues/459).

**New command** `zw`: Only matches files/folders which are a sub directory of the current directory

**New Behavior for** `zoxide query`: When the first argument is a `.`, it behaves as the `zw` command and only matches sub directories. If the argument is `~` is behaves as normal.

**TODO**: Only match directories starting in `$HOME` when first argument is a `~`


Blonteractor:
The Ubuntu build is failing, something to do with nix. No clue how to deal with it, any help? @ajeetdsouza

azaleacolburn:
@ajeetdsouza This functionality is a more specific form of the feature requested in #944. A `--baseline` flag for `zoxide query` would be a more generalization (it allows you to specify queries within specific directories outside of your own). It also solves the multiple `src` directories problem that issues such as #929 bring up.

ajeetdsouza:
Merged in https://github.com/ajeetdsouza/zoxide/pull/1027.

PR #504, opened 2022-12-18
Fix typo in README.md about nushell version compat

daniel-vainsencher:
Fixed on basis of comment in `zoxide init nushell` output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ source ~/.zoxide.nu
 ```
 
 > **Note**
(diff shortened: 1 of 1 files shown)


daniel-vainsencher:
The hint to me that something is awry was when I tried to update nu to 0.73 and found the latest github tag is 0.72...

ajeetdsouza:
Nushell v0.73.0 is released now, so I think this can be safely closed.

PR #564, opened 2023-05-08
Add instructions for importing CWDs history in Nushell 

maxim-uvarov:


KamilKleina:
Thanks for it, didn't know that, very usefull.

ajeetdsouza:
Nushell, by default, does not use SQLite for shell history. I think this is likely to help very few users, so I'm somewhat reluctant to add it to the README.

Since we do have a lot of Nushell users, if SQLite ever becomes the default, I would be willing to add a `zoxide import nushell` command, but I'm closing this till that happens. Thanks!

Issue #218, opened 2021-06-01
Support for cmder


fabioz:
Usually I use cmder as a shell which mostly wraps `cmd` in Windows -- and I'm really happy with it -- i.e.: https://cmder.net/ (https://github.com/cmderdev/cmder).

Is it possible to use zoxide in it? I only found instructions for power shell in windows, not really cmd, but maybe I missed something...

ajeetdsouza:
Hey @fabioz, cmd.exe is not supported at the moment, you'd only be able to use cmder with the Bash and PowerShell modes. I don't know cmd.exe, so I can't be very helpful here -- but if you're proficient with it, it shouldn't take very long to implement:

- `z` calls `zoxide query`
- `cd` calls `zoxide add`
- `zoxide init <SHELL>` is a useful reference for how this is implemented in other shells.

Unless this issue gains a lot of traction, I'd be hesitant to support it officially:

- Microsoft is moving everything over to PowerShell as the new default
- cmd.exe has been deprecated for years now

However, if you do end up implementing it on your own, do post it here -- I'm sure other cmd.exe users would find it helpful.

mwaitzman:
I use cmd too and I would like support for zoxide. Powershell is terrible.

rashil2000:
@ajeetdsouza FWIW, cmd on its own isn't extensible/configurable at all. Cmder uses a framework called [Clink](https://github.com/chrisant996/clink) which adds Bash's Readline support to it and makes it highly extensible (there's a comprehensive documentation available [here](https://chrisant996.github.io/clink/clink.html)). 

However, if `z` and `cd` are the only two commands/functions as you specified above, then they can be easily added through doskey aliases:
- `doskey z=zoxide query $*`
- `doskey cd=zoxide add $1 $T cd $1`

shunsambongi:
Here's a lua script I wrote for cmd.exe with clink. Put this into a `zoxide.lua` file in clink's [lua scripts location](https://chrisant996.github.io/clink/clink.html#location-of-lua-scripts). For cmder you can put it in `%cmder_root%\config\zoxide.lua`.

I only did some manual testing but it should be able to support many of zoxide's features including: 

* `zoxide init` options `--cmd`, `--hook`, and `--no-alias` (using [clink settings](https://chrisant996.github.io/clink/clink.html#clinksettings))
* `_ZO_ECHO` environment variable
* fzf integration via `zi` command

```lua
-- =============================================================================
--
[...]

PR #567, opened 2023-05-10
Add support for `cmd.exe` (experimental)

mataha:
See #218.

This PR adds experimental support for `cmd.exe` using the following Batch techniques:

- DOSKEY macros
- `for /f` with disappearing carets
- "smart" percent characters

Most of the work was done in a0c6535 - `%` characters have different meanings depending on the context in which the code is interpreted (command-line vs script), thus the beginning of the template file introduces a macro used to escape them regardless of the situation - even in `for` statements.

While working on this I had to fix path normalization on Windows in bc947f4, i.e. patching drive letter prefixes to always be uppercase (as that's how the underlying API expects them to be, even though `cmd.exe` will happily return an unnormalized component from volume path environment variable).  A similar problem [was encountered in rust-analyzer](https://github.com/rust-lang/rust-analyzer/issues/14683), which made me provide [a fix](https://github.com/rust-lang/rust-analyzer/pull/14799) there. To offset the potential overhead I've bumped `dunce` to 1.0.4 in 2fdde1c, though most of the changes there are related to licensing and packaging.

To test this, `zoxide init cmd` is piped into `cmd.exe` - anything on stderr means failure, e.g.:

```
')' is not recognized as an internal or external command,
operable program or batch file.
```


shell.nix
```diff
@@ -12,6 +12,7 @@ in pkgs.mkShell {
 
     # Shells
     pkgs.bash
+    pkgs.cmd
```
ajeetdsouza, line 15: Nix is a Linux package manager used for building the CI environment, it doesn't have `cmd.exe` in it.

src/shell.rs
```diff
@@ -93,6 +94,22 @@ mod tests {
             .stderr("");
     }
 
+    #[apply(opts)]
```
ajeetdsouza, line 97: Use `#[cfg(windows)]` for this and `#[cfg(unix)]` for the others.

shell.nix
mataha: My apologies, brainfart.

src/shell.rs
mataha: Done.

src/shell.rs
```diff
@@ -94,6 +98,22 @@ mod tests {
     }
 
     #[apply(opts)]
+    #[cfg(windows)]
+    fn cmd_cmd(cmd: Option<&str>, hook: InitHook, echo: bool, resolve_symlinks: bool) {
```
mataha, line 102: This won't run as `#[cfg(feature = "nix-dev")]` is in effect. Should I pull it to a separate module?

mataha:
It is ready.

ajeetdsouza:
This is awesome! I'm learning Batch as I review this, so forgive the intermittent comments.

README.md
```diff
@@ -180,6 +180,20 @@ zoxide can be installed in 4 easy steps:
...
+   <summary>`cmd.exe`</summary>
+
+   > Add this to the **end** of your config file or AutoRun command:
+   >
+   > ```batchfile
+   > zoxide init cmd | cmd /d >nul
+   > ```
+   >
+   > **Note**
+   > zoxide only supports `cmd.exe` with Command Extensions v2 and above.
```
ajeetdsouza, line 193: It seems Command Extensions is enabled by default on later versions of Windows. Can we remove this from the README, and add a check for it instead in the script? We could throw an informative error here, like:

```
zoxide: requires CommandExtensions v2 to run on cmd.exe, please see the instructions at <URL>
```

templates/cmd.txt
```diff
@@ -0,0 +1,161 @@
...
+{%- else -%}
+{%- call batch::indent(tabs) %}    %__builtin_pwd%^
+{%- endif -%}
+{%- call batch::indent(tabs) %})
+{%- endmacro pwd -%}
+
+{%- let section = "@rem ==========================================================================\n@rem" -%}
+{%- let not_configured = "@rem -- not configured --" -%}
+
+@echo off & setlocal DisableDelayedExpansion EnableExtensions
```
ajeetdsouza, line 35: Is it possible to reset these to their original values after initialization is done? Or do we need it to be enabled throughout?

mataha: If the contents of `zoxide init cmd` were to be run as a script, then `endlocal` take care of that. Command-echoing is per-file, so it will be reset naturally at EOF (or sooner if e.g. terminated via `exit`).

In other case (`zoxide init cmd | cmd ...`) the environment changes will be discarded and nothing will be propagated to the parent shell (except for DOSKEY macro definitions, which are per-executable and pertain for the whole duration of a session). `setlocal`/`endlocal` don't work outside of scripts anyway.

README.md
mataha: Sure.

PR #587, opened 2023-06-30
Update Nushell env update syntax

jntrnr:
Greetings!

The Nushell team is planning to move away from `let-env FOO = ...` to the more straightforward (and more honest) `$env.FOO = ...`.

This PR should update zoxide to use the supported environment update syntax.

### DIFF
--- a/templates/nushell.txt
+++ b/templates/nushell.txt
@@ -13,20 +13,20 @@
 {%- else -%}
 # Initialize hook to add new entries to the database.
 if (not ($env | default false __zoxide_hooked | get __zoxide_hooked)) {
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Thank you!

happysalada:
@ajeetdsouza would it be possible to make a release with this fix applied ?
for package maintainers, we either have to manually apply this patch or wait for a release.
Both are not a problem, I just thought I might ask anyway.

ajeetdsouza:
Just saw that this was officially removed in 0.83.1. I've been swamped with work, but I'll try to cut a release this weekend.

PR #590, opened 2023-07-05
New installer options

aarondill:
This adds support for serveral new command line options (taken from the usage):
```
    -b, --bin-dir
                Override the bin installation directory [default: /home/aaron/.local/bin]

        -m, --man-dir
                Override the man installation directory [default: /home/aaron/.local/share/man]

        -a, --arch
                Override the architecture identified by the installer [default: x86_64-unknown-linux-musl]

        -s, --sudo
                Override the command used to elevate to root privaliges [default: sudo/doas]

        -h, --help
                Display this help message
```
This also (coincidentally) allows for $BIN_DIR, $MAN_DIR, $ARCH, and $SUDO to be set to create default settings for these.
These variables are *always* overwritten by the command line arguments.

This diff is unusually large, as my formatter ([shfmt](https://github.com/mvdan/sh)) ran, reformatting the majority of the script.

This also adds a few utilities to the script (`log`/`abort`) and modifies `err` to simply output to stderr.
This allows for more fine grain control over script flow and output, without relying on `echo >&2`.

Each commit should include descriptions regarding their necessity. If you don't understand something or just want some extra clarity, I'm happy to explain any of my choices.

## NOTE:

These options are **not** yet fully tested. I am posting this here while I test on my own machine for the purposes of a second set of eyes, but this should *not* be merged until tested by at least two people :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ zoxide can be installed in 4 easy steps:
    > The recommended way to install zoxide is via the install script:
    >
    > ```sh
-   > curl -sS https://raw.githubusercontent.com/ajeetdsouza/zoxide/main/install.sh | bash
+   > curl -sSfL https://raw.githubusercontent.com/ajeetdsouza/zoxide/main/install.sh | sh
    > ```
    >
    > Or, you can use a package manager:
@@ -109,7 +109,7 @@ zoxide can be installed in 4 easy steps:
    > Or, run this command in your terminal:
    >
    > ```sh
-   > curl -sS https://raw.githubusercontent.com/ajeetdsouza/zoxide/main/install.sh | bash
+   > curl -sSfL https://raw.githubusercontent.com/ajeetdsouza/zoxide/main/install.sh | sh
    > ```
 
    </details>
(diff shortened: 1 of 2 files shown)


aarondill:
one known issue, which might be complicated to resolve, is that the use of `sudo` is based on the writability of `_bin_dir`, so if `_man_dir` is under "$HOME", this could result in files owned by root in the user's home directory.

aarondill:
> one known issue, which might be complicated to resolve, is that the use of `sudo` is based on the writability of `_bin_dir`, so if `_man_dir` is under "$HOME", this could result in files owned by root in the user's home directory.

Additionally, if `$_bin_dir` is writable, but `$_man_dir` is *not*, then the script fails during man page installation. I think the only ways to resolve this is to reset the `$_sudo` variable before installing man pages, or to create two `$_sudo` variables (ie `$_sudo`/`$_sudo_man`)

aarondill:
I *have* now tested all the options (and they work *on my system*), and I have fixed the above mentioned error about _man_dir and _bin_dir permission mismatches. 

@ajeetdsouza, Please take a look and let me know what you think. I would also appreciate if you could test these on your own system to avoid "it works on my machine" :smile:

hanoii:
For what it worth, I am doing:

`curl -sS https://raw.githubusercontent.com/ajeetdsouza/zoxide/e3ab444976c49a518469bcc30c74b6780cd33fde/install.sh | HOME=/opt/zoxide bash && cp -R /opt/zoxide/.local/* /usr/local && rm -fr /opt/zoxide` as a workaound to move the installed files to a directory of choice.

Issue #595, opened 2023-07-26
Error with nushell 0.83.0


18183883296:
The project is not well-adapted to the latest version of nushell, version 0.83.0, and it displays the following error:
```
Error: nu::parser::input_type_mismatch

  × Command does not support string input.
    ╭─[C:\Users\Sunrise\.zoxide.nu:27:1]
 27 │   let arg0 = ($rest | append '~').0
 28 │   let path = if (($rest | length) <= 1) and ($arg0 == '-' or ($arg0 | path expand | path type) == dir) {
    ·                           ───┬──
    ·                              ╰── command doesn't support string input
 29 │     $arg0
    ╰────
```

asim215:
Solution is following:
`($rest | str length).0 <= 1`
I also changed everywhere:
from `let-env x` to `$env.x`

```
# Initialize hook to add new entries to the database.
if (not ($env | default false __zoxide_hooked | get __zoxide_hooked)) {
  $env.__zoxide_hooked = true
  $env.config = ($env | default {} config).config
  $env.config = ($env.config | default {} hooks)
  $env.config = ($env.config | update hooks ($env.config.hooks | default {} env_change))
  $env.config = ($env.config | update hooks.env_change ($env.config.hooks.env_change | default [] PWD))
  $env.config = ($env.config | update hooks.env_change.PWD ($env.config.hooks.env_change.PWD | append {|_, dir|
    zoxide add -- $dir
  }))
}

# =============================================================================
#
# When using zoxide with --no-cmd, alias these internal functions as desired.
#

# Jump to a directory using only keywords.
def-env __zoxide_z [...rest:string] {
  let arg0 = ($rest | append '~').0
  let path = if (($rest | str length).0 <= 1) and ($arg0 == '-' or ($arg0 | path expand | path type) == dir) {
    $arg0
  } else {
    (zoxide query --exclude $env.PWD -- $rest | str trim -r -c "\n")
  }
  cd $path
}

# Jump to a directory using interactive search.
def-env __zoxide_zi  [...rest:string] {
  cd $'(zoxide query --interactive -- $rest | str trim -r -c "\n")'
}

# =============================================================================
#
# Commands for zoxide. Disable these using --no-cmd.
#

alias z = __zoxide_z
alias zi = __zoxide_zi
```

mati865:
@asim215  this is not a fix because it changes the way zoxide works.
[...]

PR #596, opened 2023-07-26
fix: use `str length` for str instead of `length`

mrdgo:
nushell won't start as long as you use `length`.

```
[1 2 3 4] | length
```
works.

```
"hello world" | str length
```
works.

Different permutations don't.

### DIFF
--- a/templates/nushell.txt
+++ b/templates/nushell.txt
@@ -41,7 +41,7 @@ if (not ($env | default false __zoxide_hooked | get __zoxide_hooked)) {
 # Jump to a directory using only keywords.
 def-env __zoxide_z [...rest:string] {
   let arg0 = ($rest | append '~').0
-  let path = if (($rest | length) <= 1) and ($arg0 == '-' or ($arg0 | path expand | path type) == dir) {
+  let path = if (($rest | str length) <= 1) and ($arg0 == '-' or ($arg0 | path expand | path type) == dir) {
     $arg0
   } else {
     (zoxide query --exclude $env.PWD -- $rest | str trim -r -c "\n")
--- a/templates/nushell.txt
+++ b/templates/nushell.txt
@@ -39,9 +39,10 @@ if (not ($env | default false __zoxide_hooked | get __zoxide_hooked)) {
 #
 
 # Jump to a directory using only keywords.
-def-env __zoxide_z [...rest:string] {
+# def-env __zoxide_z [...rest:string] {
+def-env __zoxide_z [...rest] {
   let arg0 = ($rest | append '~').0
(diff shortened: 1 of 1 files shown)


mati865:
This won't work, I have tried the same but it fails when trying to use zoxide:
```
❯ z target
Error: nu::shell::type_mismatch

  × Type mismatch during operation.
    ╭─[/home/mateusz/.zoxide.nu:27:1]
 27 │   let arg0 = ($rest | append '~').0
 28 │   let path = if (($rest | str length) <= 1) and ($arg0 == '-' or ($arg0 | path expand | path type) == dir) {
    ·                  ──────────┬───────── ─┬ ┬
    ·                            │           │ ╰── int
    ·                            │           ╰── type mismatch for operator
    ·                            ╰── list<int>
 29 │     $arg0
    ╰────
```

maxim-uvarov:
@mati865 I believe this is a bug of nushell 0.83.0 (check it here https://github.com/nushell/nushell/issues/9809)

Otherwise it should work.

fdncred:
A hack for this, until we fix the bug, to get people up and running is changing
```nushell
def-env __zoxide_z [...rest:string]
```
to
```nushell
def-env __zoxide_z [...rest]
```
and do the same for `__zoxide_zi`

mati865:
> @mati865 I believe this is a bug of nushell 0.83.0 (check it here https://github.com/nushell/nushell/issues/9809)
> 
> Otherwise it should work.
> 
> 

Yup, that's why I have mentioned it in referenced item.

fdncred:
fwiw, I'd run with the "temporary fix" for now. I'm not sure how long it'll be to fix it the "right" way. We can do another PR here when `...rest: string` works again.

asim215:
I fix it by another way. 
See #595 .

amtoine:
hello there from Nushell :wave: :yum: 

so yeah `str length` is not a drop in replacement for `length` in that case, because `$rest` is not a `string` but a `list<string>`, which can be seen with a `describe` as in https://github.com/nushell/nushell/issues/9720#issuecomment-1640551042 :+1: 

this was a bug with how `$rest` was passed around and has been fixed upstream in https://github.com/nushell/nushell/pull/9816 :relieved: 
should be available soon in a patch release :blush:

ajeetdsouza:
That's good news! In that case, I think it would be better to wait for the patch release rather than release a hack here.

amtoine:
> That's good news! In that case, I think it would be better to wait for the patch release rather than release a hack here.
>
we're working on it... :recycle: 
i'd say the best and safest is to keep those kind of PRs opened and ready, just in case, and all stay in touch. how's that sound? :innocent:

ajeetdsouza:
This seems to be fixed in Nushell, so I'm closing this. Thanks!

PR #597, opened 2023-07-27
add global to all powershell functions

matan-h:
This will allow loading zoxide from a powershell function.
For example, when I try to execute `z` with powershell profile like this:
```powershell
function load_zoxide(){
      invoke-Expression (& { (zoxide init powershell | Out-String) })
}
load_zoxide;
``` 
It results with `The term '__zoxide_z' is not recognized`, because `z` alias is global, but `__zoxide_z` function is not.

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -83,7 +83,7 @@ function global:__zoxide_hook {
 $global:__zoxide_hooked = (Get-Variable __zoxide_hooked -ErrorAction SilentlyContinue -ValueOnly)
 if ($global:__zoxide_hooked -ne 1) {
     $global:__zoxide_hooked = 1
-    $__zoxide_prompt_old = $function:prompt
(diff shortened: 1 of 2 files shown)


templates/powershell.txt
```diff
@@ -85,7 +85,7 @@ if ($__zoxide_hooked -ne 1) {
     $__zoxide_hooked = 1
```
ajeetdsouza, line 85: Shouldn't `__zoxide_hooked` also be marked as `global`?

matan-h: Ok, I mark it also as global. Currently, in my profile, I use str.replace: `(zoxide init powershell | Out-String).Replace("function __","function global:__")`. And it works, so I am not sure what is the impact of this variable

ajeetdsouza:
This is released, thanks!

PR #619, opened 2023-09-11
Specify exact winget id

connorjs:
Specifies exact winget identifier

---

Reasoning: I have a wrapper `winget-install` script that enforces a consistent location. It also uses `-e` to only install exact matches. So, I figured this README should show the exact ID.

Feel free to reject/close if you prefer just `zoxide`.


ajeetdsouza:
Thanks!

PR #631, opened 2023-10-14
Remove creation of file z! on init in fish

arsenier:
In fish 3.1 executing `zoxide init fish | source` creates file `z!` in working directory. Some experimentation showed, that such commands: `echo ^(echo 'z!')` have similar effect.
Caret removal fixes issue with creating files and does not affect init behaivour:
```
$ cat ~/.config/fish/zoxide_init | source
$ echo $__zoxide_z_prefix_regex
z!
```
(zoxide_init is changed init script)

### DIFF
--- a/templates/fish.txt
+++ b/templates/fish.txt
@@ -65,7 +65,7 @@ end
 if test -z $__zoxide_z_prefix
     set __zoxide_z_prefix 'z!'
 end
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
This seems like a bug in Fish rather than zoxide, that has been patched in the later versions. Unfortunately, this solution can cause problems, because `z!` is no longer treated like prefix:

```console
$ z az!
cd: The directory 'a' does not exist
```

Issue #630, opened 2023-10-14
nushell breaking changes


fdncred:
Heads-up @ajeetdsouza, With https://github.com/nushell/nushell/pull/10566 we will deprecate the `def-env` syntax in nushell. Our next release 0.86.0 on Oct-17 will allow both old and new syntax but the next release, 4 weeks later, [will produce a warning](https://github.com/nushell/nushell/pull/10715).  The new syntax is `def --env`.

marcelarie:
Thanks this was really helpful!

fdncred:
@lpnh This last part has nothing to do with zoxide.
```
Error:   × Config record contains invalid values or unknown settings

Error:   × Error while applying config changes
     ╭─[/home/user/.config/nushell/config.nu:152:1]
 152 │
 153 │ ╭─▶     cd: {
 154 │ │           abbreviations: false # allows `cd s/o/f` to expand to `cd some/other/folder`
 155 │ ├─▶     }
     · ╰──── $env.config.cd is an unknown config setting
 156 │
     ╰────
  help: This value will not appear in your $env.config record.
```
$env.config.cd.abbreviations was removed from nushell in favor of using our path completions that work similarly now. You just need to comment that out of your config.nu (or remove that part).

acidghost:
As far as I can see the only two places where zoxide needs to be updated are https://github.com/ajeetdsouza/zoxide/blob/073500cbdf8ba211e5115fc6e51411cf2468fc5b/templates/nushell.txt#L42-L61

I think it's just a matter of replacing `def-env` with `def --env`.

Mickychen00:
Hello. I have notice there is a fix commit "f537a4e6d2f8c2eb84c63f79e290a6d1b16eeb71" in zoxide. How can I install this dev version of zoxide?

marcelarie:
Not sure but it might be `brew install --HEAD zoxide` in macOS

Slushee-a:
When will this be pushed to crates.io? Currently having to install from git.

Aarkon:
Nushell has just published their update, removing `def-env` altogether. As far as I understand, this breaks zoxide entirely in nu. Providing a fixed release on crates.rs anytime soon has become somewhat urgent now, I’d say.

dtscherer:
For now I'm just using this in my env.nu
```Nushell
zoxide init nushell | str replace "def-env" "def --env" --all | save -f ~/.config/nushell/modules/.zoxide.nu
```

[...]

PR #632, opened 2023-10-19
Update nushell.txt to fit nushell

hackers267:
Change the `def-env` to `def --env` to fit the nushell v8.8, because the `def-env` is deprecated and be removed in v8.8

Fixes https://github.com/ajeetdsouza/zoxide/issues/630

PR #642, opened 2023-11-14
Make zoxide compitable with upcoming nushell release

WindSoilder:
nushell will deprecate `def-env` and use `def --env` in the next release(0.87)

Currently `def --env` is already available in latest version (0.86)

Relative: #630


WindSoilder:
Oh, sorry that a similar pr is already exists: #632 
I'll close this pr.

Issue #637, opened 2023-11-02
Add Webi to list of install methods again.


coolaj86:
Hey @ajeetdsouza,

Would you accept a PR to add Webi back to the install methods?

I believe that over the years we've been able to find solutions for all of the most important philosophical issues and edge case bugs that affected the installer.

If there's anything outstanding, I'd love to hear about it.

ajeetdsouza:
Hey @coolaj86, I appreciate the work you've done on `webi`. My initial motivation for switching over to Webi was to ease maintenance, but it ended up doing the opposite:

- Issues that were reported now took much longer to resolve, because they now relied on me getting time to figure out the issue with `webi`, create a PR, and it getting reviewed / merged there.
- There were some philosophical issues I had reported right at the start (`webi` modifying config files, installing itself onto your computer without warning, etc.). I had assumed that this would not trouble too many people, but quite a few users expressed discontent with these changes. I'm not sure if this has been fixed yet.
- `webi`, being an actively developed repository, had bugs that broke the installer a couple of times. While this is inevitable, it does mean that it involves a time investment from me every time it breaks - OTOH, the current zoxide installer has not changed since it was (re) added.

I may reconsider this in future, but for now, I'd prefer to stick with the current installation script. Hope you understand!

coolaj86:
I wasn't asking you to change it to the default.

I'm just asking about putting it in the table with the others like snap, choco,  asdf, etc.

PR #647, opened 2023-11-23
doc: add Webi to installers table

coolaj86:
Re: #637 

- [x] macOS
- [x] Linuxes
- [x] Windows
- [ ] BSDs (being worked on)
- [ ] Android (being worked on)


ajeetdsouza:
We've tried switching to Webi before, but the maintenance burden of fixing bugs in another repository and following up till it gets merged is more than just maintaining an in-house installer. Our current installer has been working perfectly for years without complaint, so I think we should keep using it as-is.

PR #655, opened 2023-12-08
Add limit option for zoxide query

shouya:
This PR adds a simple option to specify the number of directories to output in `zoxide query`.

This option is useful when you only want to peek what the matching directories look like without listing them all. I'm currently expecting to use it in a hand-crafted fish completion script.

I currently write the completion script as:

```fish
complete -c z -x -a '(zoxide query -l -- $argv 2>/dev/null | head -n 20)'
```

The redirection is needed because otherwise you get a panic due to EPIPE. Now with this limit option I can just write:

```fish
complete -c z -x -a '(zoxide query -l -n 20 $argv)'
```

### DIFF
--- a/contrib/completions/_zoxide
+++ b/contrib/completions/_zoxide
@@ -117,6 +117,8 @@ _arguments "${_arguments_options[@]}" \
 ;;
 (query)
 _arguments "${_arguments_options[@]}" \
+'-n+[Limit number of results in List mode]:LIMIT: ' \
+'--limit=[Limit number of results in List mode]:LIMIT: ' \
 '--exclude=[Exclude the current directory]:path:_files -/' \
 '-a[Show unavailable directories]' \
 '--all[Show unavailable directories]' \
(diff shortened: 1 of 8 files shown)


shouya:
Weirdly I can no longer reproduce the error, even though I still run the same version when I met the error.

From strace I can still see `SIGPIPE` is sent to the process, but the process no longer crashes.

``` shell
$ strace -e 'trace=!all' zoxide query -l | head -n 1
/home/shou/tmp
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=326529, si_uid=1000} ---
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=326529, si_uid=1000} ---
+++ exited with 0 +++
```

Then I noticed the [BrokenPipeHandler](https://github.com/ajeetdsouza/zoxide/blob/e55e0b27e27930ab8af81999a91b56cf0cd6ff6f/src/error.rs#L18) is already in-place two years ago, long before I attempt to fix the issue. I'm puzzled now on what went on back when I submit the PR.

But anyway, since we no longer see the panic and the stream is also efficiently consumed with `head`, I see no reason to add the limiting feature. Feel free to close this PR if you agree.

ajeetdsouza:
Strange. Perhaps you had an older version of zoxide installed? Either way, I'll close the PR for now. Do let me know if it crops up again.

PR #663, opened 2024-01-10
Update nushell template to use rest operator where needed

### LINKED ISSUE #662: Nushell 0.89 deprecation
Context: since [nushell 0.89](https://www.nushell.sh/blog/2024-01-09-nushell_0_89_0.html), the `...` operator can be applied to params passed to a command; and it is deprecated not to do so (giving the following output when using zoxide):
```
Error:   × Automatically spreading lists is deprecated
    ╭─[/home/zuruh/dotfiles/.config/nushell/plugins/zoxide.nu:30:1]
 30 │   } else {
 31 │     (zoxide query --exclude $env.PWD -- $rest | str trim -r -c "\n")
    ·                                         ──┬──
    ·                                           ╰── Spreading lists automatically when calling external commands is deprecated and will be removed in 0.91.
 32 │   }
    ╰────
  help: Use the spread operator (put a '...' before the argument)
```

Zuruuh:
Close #662, tested locally and there's no deprecation anymore with this version :tada:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Nushell: support for v0.86.0.
+- Nushell: support for v0.89.0.
 
 ## [0.9.2] - 2023-08-04
 
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ zoxide can be installed in 4 easy steps:
    > ```
    >
    > **Note**
-   > zoxide only supports Nushell v0.86.0+.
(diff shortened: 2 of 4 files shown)


dywedir:
Perhaps it would also be reasonable to change the minimum supported version of Nushell to v0.89.0?
I see mentions in `README.md`, `man/man1/zoxide-init.1` and finally at the bottom of `templates/nushell.txt`

poliorcetics:
Nushell 0.90 is just out and 0.89 has been the most downloaded version for a few weeks on crates.io now so only the changes proposed by @dywedir are left and updating the minimum nushell version should not impact people negatively :)

ajeetdsouza:
Thanks!

PR #680, opened 2024-02-06
fix command for powershell

AbhimanyuAryan:
the old command doesn't work and gives error

```
zoxide: PWD hooks are not supported below powershell 6.
        Use 'zoxide init powershell --hook prompt' instead.
At line:1 char:1
+ . 'C:\Users\aryan\Documents\WindowsPowerShell\Microsoft.PowerShell_pr ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Microsoft.PowerShell_profile.ps1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ zoxide can be installed in 4 easy steps:
    > `echo $profile` in PowerShell):
    >
    > ```powershell
-   > Invoke-Expression (& { (zoxide init powershell | Out-String) })
+   > Invoke-Expression (& { (zoxide init powershell --hook prompt | Out-String) })
    > ```
 
    </details>


ajeetdsouza:
PWD hooks work on PowerShell 7+ and above, the prompt hook has been kept as a fallback for older versions. You're probably running PowerShell 5, you can either upgrade it or modify the command to `--hook prompt`.

PR #695, opened 2024-02-14
support autocd option

solodov:
In bash, when autocd option is set and user enters a directory name as a command, it results in a very specific call to cd:

cd -- [directory name]

zoxide's directory changing function passes all arguments to __zoxide_z as is, including the "--" first argument. By detecting this and skipping the first argument changing directory works with autocd set.

This particular bug only happens when initializing zoxide with `--cmd cd`. Here's how it breaks in bash:

```bash
$ shopt -c autocd
$ eval $(zoxide init --cmd cd bash)"
$ ..
cd -- ..
zoxide: no match found
```

### DIFF
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -75,6 +75,8 @@ function __zoxide_z() {
         __zoxide_cd "${OLDPWD}"
     elif [[ $# -eq 1 && -d $1 ]]; then
         __zoxide_cd "$1"
+    elif [[ $# -eq 2 && $1 == '--' ]]; then
+        __zoxide_cd "$2"
     elif [[ ${@: -1} == "${__zoxide_z_prefix}"?* ]]; then
         # shellcheck disable=SC2124
         \builtin local result="${@: -1}"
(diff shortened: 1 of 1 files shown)


PhoenixmitX:
would love to see this merge

solodov:
As far as I can tell this PR is ready for merging. Let me know if there's anything missing.

german-moreno283:
Would this work for zsh as well?

solodov:
> Would this work for zsh as well?

I updated initial comment with the reproduction steps for bash. Checked in zsh and seems to work correctly already:

```zsh
% setopt  autocd
% eval "$(zoxide init --cmd cd zsh)"
% /
```

charlesbvll:
@ajeetdsouza Is there any update on this? Seems like a very useful feature to have!

solodov:
> @ajeetdsouza Is there any update on this? Seems like a very useful feature to have!

PR is ready to be merged, as far as I'm concerned. I've been using this fix for a few months now, everything seems to be ok.

fwam:
bump

tonyOehme:
This seems like a neat improvement. @ajeetdsouza any updates on this?

ajeetdsouza:
Thank you, and sorry about the delay!

PR #755, opened 2024-03-13
doc: Documents the exact command needed

ActuallyHappening:
This is now easier to install for nushells


ajeetdsouza:
This may be personal bias, but I don't usually like scripts that modify my config files - I prefer to change them myself. Thanks all the same!

ActuallyHappening:
All good

PR #777, opened 2024-04-01
fix resolves symlink on `cd` at zsh

### LINKED ISSUE #740: Case-sensitivity issues with zoxide database?
I'm using MacOS and  have a `~/Code` directory, and am finding inconsistent results when trying to change directory using `z co` vs `z code`...

CleanShot 2024-02-29 at 21 23 29

But then if I `zoxide edit` to look at database, I see it's database them as SEPARATE folders, when only one folder with this name exists?

CleanShot 2024-02-29 at 21 25 48

Is there a way to prevent zoxide from doing this?

Note: I'm coming from rupa/z, where the only thing that gets indexed is `~/Code` (with capital `C`), therefore both `z co` and `z code` both bring me to `~/Code` (with capital `C`), as I'd expect.

dalton-oliveira:
### Description

Having `_ZO_RESOLVE_SYMLINKS` env var defined as "1", makes zoxide to chase symink of directory before adding it to  the database. This behavior can be more coherent if the `cd` command also does the same. That's specially useful for APFS which is case insensitive by default. Closes #740.

### What was done

Adds `CHASE_LINKS` option to `cd`, the same one used in `pwd` on zsh.


jesseleite:
Ooo, not sure how to test locally since I've brew installed, but excited for this one! 🎉

dalton-oliveira:
@ajeetdsouza thank you for this amazing tool! It has already saving me a lot of keystrokes. Hope this PR is adequate and thanks for taking a look.

jesseleite:
Tested this branch using @dalton-oliveira's instructions [here](https://github.com/ajeetdsouza/zoxide/issues/740#issuecomment-2030459007) and it seems to totally fix my issue in #740 😍

ajeetdsouza:
@dalton-oliveira just one question - why do we not use `cd -P` on Windows?

dalton-oliveira:
@ajeetdsouza I was concerned to break something that I couldn't check manually. Just removed the conditional

jesseleite:
Thanks for doing this @dalton-oliveira! ❤️

jesseleite:
Just tested again since that last commit. Still works great / fixes my MacOS case sensitivity issues in #740 👍

Thoughts on this one @ajeetdsouza 🙏

duncanmcclean:
I have a similar issue to @jesseleite - would love to see this merged! 😄

dalton-oliveira:
Hey @ajeetdsouza have just rebased with main. It's been here for a while so please feel free to reject the PR if you doesn't intent to merge it :)

jesseleite:
Thoughts on this one @ajeetdsouza? ...Would be huge little thing for us MacOS users, and it works great! 🙏🙏🙏

PS. Appreciate all you do! ❤️

austenc:
Just ran into this too, hope it gets merged soon!

PR #779, opened 2024-04-04
Update fish instruction in README.md

stephane-archer:


ajeetdsouza:
Wouldn't recommend adding this to the README. We get a number of issues from users who have not added zoxide to their PATH, and this line would make the issue even harder for them to debug.

Issue #775, opened 2024-03-29
Conflict between zoxide, zsh-autocomplete, and fzf results in erroneous behavior


argamanza:
### Environment
Shell: zsh 5.9
OS: macOS 14.1

**Tools & Versions:**
zoxide: 0.9.4
fzf: 0.48.1 (installed via brew)
zsh-autocomplete: 23.07.13

### Description
When using zoxide in conjunction with zsh-autocomplete and fzf, the expected behavior of navigating to directories sorted by frecency fails. Specifically, after initiating a directory search with `z prefix<SPACE><TAB>`, selecting a directory from the fzf menu, and pressing `<ENTER>`, the fzf menu reopens with the input "0n" instead of changing the directory.

### Steps to Reproduce
1. Configure .zshrc with the following setup:
```
# Set up fzf key bindings and fuzzy completion
eval "$(fzf --zsh)"

# Initialize zsh-autocomplete
source /opt/homebrew/share/zsh-autocomplete/zsh-autocomplete.plugin.zsh

# Initialize zoxide
eval "$(zoxide init zsh)"
```
2. Type z `sometext<SPACE><TAB>` to invoke the fzf menu.
3. Select a directory and press `<ENTER>`.

### Expected Behavior
Upon selecting a directory from the fzf menu and pressing `<ENTER>`, the shell should navigate to the selected directory.

### Actual Behavior
After selecting a directory, pressing `<ENTER>` reopens the fzf menu with `0n` pre-filled, leading to an empty list. Exiting this state requires pressing `<ESCAPE>` three times or manually correcting the input.

### Additional Information
- When using zoxide on a directory prefix with only one possible completion it will paste the command 4 times and `cd` into it.
- This issue does not occur when zsh-autocomplete is not initialized in the .zshrc file.
- using `zi` works as expected, seems like the issue is with the `<TAB>` key press.
- I will attach a video demonstrating the described behavior.

### Attempts to Resolve
Using `autoload -U compinit; compinit` as demonstrated below, instead of the zsh-autocomplete initialization allows zoxide to function as expected but without the desired autocomplete features.
```
autoload -U compinit; compinit

# Initialize zsh-autocomplete
source /opt/homebrew/share/zsh-autocomplete/zsh-autocomplete.plugin.zsh

# Initialize zoxide
eval "$(zoxide init zsh)"
```

## TL;DR
[...]

PR #785, opened 2024-04-09
fix: Space-Tab interactive completions repeating

### LINKED ISSUE #780: Interactive expansion doesn't work as expected when it matches a single entry then echos command 3 times
This was working, and no longer is. I'm still not a daily driver of zoxide, so can't report on when exactly it stopped working. I did do some upgrades of packages recently (using brew), so could be an incompatibility of new versions? Not sure what else could be causing the problem.

When using `z path<SPACE><TAB>` it cd's into that directory, and then echos the current directory:
SCR-20240405-khjf

Here are what are likely relevant package versions currently installed:
```
fzf 0.49.0
zoxide 0.9.4
zsh-autocomplete 23.07.13
zsh 5.9 (x86_64-apple-darwin23.0)
```

I'm using `oh-my-zsh` but not any plugins, i've got auto-complete being loaded in my `.zshrc` file, as below (in order)

```
source $ZSH/oh-my-zsh.sh
source $(brew --prefix)/share/zsh-autocomplete/zsh-autocomplete.plugin.zsh

eval "$(fzf --zsh)"
# Other auto-completions
eval "$(zoxide init zsh)"
```

Any ideas? Is this a bug? I can see that fzf is still working in my shell.

ashleybartlett:
Closes #780 and #775

Looks you needed to still call `compadd` to set a completion for the current function call, otherwise it kept trying to re-run the completion. 

This made it look like `Enter` wasn't accepting results in fzf (you could hit `Enter` 3 times and it would eventually work), and the single match repeated 3 times, as I suspect there is some completion retry if it fails, but didn't look for that answer.

I've left the clearing of the results in, as I think it might be good hygiene.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- zsh: Space-Tab completion repeating output multiple times when matching single directory in history
 - fish: detect infinite loop when using `alias cd=z`.
 - fish / Nushell / PowerShell: handle queries that look like args (e.g. `z -x`).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


ajeetdsouza:
Thanks!

Issue #513, opened 2023-01-13
Improve default Zsh completion for z


gi1242:
By default if I type

gi1242:
(Apologies; when I initially submitted my report my comment got deleted except for the first thre words. I edited my comment to include the fill report. Sorry if you got an almost blank issue report email notification.)

Percee:
The change you're proposing isn't working for me. Instead of doing `eval "$(zoxide init zsh)"`  I've pasted the output of `zoxide init zsh` in my zshrc and modified `_files -/` to `_cd`, but this doesn't fix it.

I want to complete the `hello_world` directory by doing `z hwo<TAB>` but that doesn't work.

OrbitalHustler:
Thanks @ryankask ! Figured out my issue, I had to put my autoload -Uz compinit first, eval the updated function next, and finish by compinit -i -u

@ajeetdsouza any chance of this being added into zoxide? do you see an issue with this fix?

nubunto:
still doesn't work for me, not on my personal machine but it does work on my work one. both are MacOS, with identical config.

using `z` works as expected

sevmonster:
I've been doing this for the past while:
```zsh
eval "${$(zoxide init zsh):s#_files -/#_cd#}"                               alias cd=z   
```

ijcd:
I took a different approach: rather than making `z` complete like `cd`, I made it complete from the zoxide frecency database directly.
                                                                                                                                                                        
My reasoning is that if I want `cd`-style completion, I'll just use `cd`. The whole point of `z` is frecency-based jumping, so why not have completion match that?

This function allows me to use <tab> to get the zoxide database values for completion, but still use <enter> to jump into the fuzzy picker. Feels like the best of both to me.
                                                                                                                                                                        
Here's the completion function I'm using for `z`, `zi`, and a custom `j` alias:

```zsh
# Tab completion for zoxide - completes from frecency database instead of local dirs
function _zoxide_complete() {
    # Only complete when cursor is at end of line
    [[ "${#words[@]}" -eq "${CURRENT}" ]] || return 0

    local -a dirs expl
    # Query zoxide db - works with or without args (empty args = all entries)
    dirs=(${(f)"$(zoxide query -l -- ${words[2,-1]} 2>/dev/null | head -15)"})

    if (( ${#dirs} )); then
        # Force menu selection (no common-prefix completion)
        compstate[insert]=menu
        compstate[list]=list
[...]

PR #787, opened 2024-04-11
zsh: better local directory completion

gi1242:
Fixes issue #513

### DIFF
--- a/templates/zsh.txt
+++ b/templates/zsh.txt
@@ -110,11 +110,11 @@ if [[ -o zle ]]; then
             \builtin bindkey '\e[0n' '__zoxide_z_complete_helper'
             # Send '\e[0n' to console input.
             \builtin printf '\e[5n'
+        fi
 
 	    # Report that the completion was successful, so that we don't fall back
 	    # to another completion function.
 	    return 0
-        fi
     }
 
     function __zoxide_z_complete_helper() {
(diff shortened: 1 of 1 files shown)


templates/zsh.txt
```diff
@@ -98,7 +98,9 @@ if [[ -o zle ]]; then
 
         if [[ "{{ "${#words[@]}" }}" -eq 2 ]]; then
             # Show completions for local directories.
-            _files -/
+            _cd -/
+
+	    # Don't return 0 so we can fall back to another comleter if unsuccessful.
```
ajeetdsouza, line 103: Avoid this - we don't want to fallback to other completions.

ajeetdsouza:
Merged. Thank you!

templates/zsh.txt
gi1242: We want this here @ajeetdsouza 

Without it local directory completion will work very badly. For instance 

    cd oo<tab>

will not complete `foo`, and instead complete nothing.

ajeetdsouza: Won't it also complete things like file names, which we don't want?

gi1242: Zsh fallback is smart :). It first tries completing corrections as directories; if that fails it completes files.

Also, looking at the logic -- the way I had the patch originally `z o<tab>` will fall back to corrections, but `z o <tab>` will not. That's probably what you want, since that's where you do the fzf / history magic. Best to keep `z o<tab>` as close to `cd o<tab>` completion as possible (which is what my patch did).

ajeetdsouza: Got it. Fixed this in `main`.

PR #802, opened 2024-04-25
Support bulk edit

phanen:


ErRusso (approved):

phanen:
close as i have used another script for a long time and wrap zoxide inside it

PR #806, opened 2024-05-01
README.md - refer to man pages

nielsbom:
Not everybody's first instinct is to `man zoxide` so let's be explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ zoxide works on all major shells.
 [Getting started](#getting-started) •
 [Installation](#installation) •
 [Configuration](#configuration) •
+[Usage](#usage) •
 [Integrations](#third-party-integrations)
 
 </div>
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Appreciate the PR, but I don't think this is necessary. The README provides all the information required to use zoxide, so most people would not need manpages. The people who tend to reach for manpages before a browser will find it on their own.

nielsbom:
My experience: I searched for 20 minutes in the Wiki and the discussions before I found out that zoxide had man pages. I wanted to list all the saved paths.

But n=1, you know zoxide's audience better 🤷

Maybe add a top-level page to [the wiki](https://github.com/ajeetdsouza/zoxide/wiki) referring to the man pages?

ajeetdsouza:
Added to the wiki: https://github.com/ajeetdsouza/zoxide/wiki

PR #818, opened 2024-05-17
fix(elvish): before-chdir should assign to oldpwd upvalue

alaviss:
elvish uses lexical scoping for closure capture, as such all `oldpwd` references within the script refers to the `oldpwd` defined at the top. However, before-chdir hook is trying to assign to `oldpwd` within the editor scope, which is not used by this script.

This manifested as a bug in which:

```
~
> mkdir -p /tmp/another

~
> z /tmp

/tmp
> z another

/tmp/another
> z -

~
> # The previous dir should be /tmp not ~!
```

Because the hook was updating a variable that was not used.

This PR fixes the hook so that before-chdir assign to the proper upvalue.


ajeetdsouza:
Thanks!

PR #854, opened 2024-07-09
Add _ZO_SEARCH_DIR environment variable

Hylian:
When set, this environment variable filters query results to only those within the specified directory.

This can be used to define a variant of `__zoxide_zi` that only searches within the current dir, with `_ZO_SEARCH_DIR="./"`.


ajeetdsouza:
Can't you just run `zoxide query "$PWD" <your query here>` for this?

Hylian:
That's... a very good point! ^_^;;

This patch originated from me not knowing zoxide had fzf integration, and my custom fzf function was reinvoking `zoxide query` as the search query updated. With the actual zoxide interactive mode, this isn't an issue.

Hylian:
FWIW, it seems like to invoke the interactive finder with all directories under pwd, you need to do: `__zoxide_zi "${PWD}" "/"`, as otherwise, it will populate the list with only the current directory.

PR #865, opened 2024-08-03
fix(fish): prefix builtin for all non-reserved words

phanen:
There are a few builtins used here are not documented as reserved words (https://fishshell.com/docs/current/cmds/function.html#description).

They are possible to be overrided by user functions, while we may depend on its default behavior.


ajeetdsouza:
Thank you!

amiorin:
It gives this error with fish 3.3.1
image

ajeetdsouza:
Fixed in v0.9.6.

PR #866, opened 2024-08-03
fix(fish): reset status code before `set`

phanen:
`set` will preserve the exit status of the last command (https://github.com/fish-shell/fish-shell/issues/3651).


ajeetdsouza:
As far as I can tell, the issue occurs when you use `set` with no subcommand (e.g. `set x 1` rather than `set x (echo 1)`). However, the `set` command in this PR does have a subcommand in it - so there shouldn't be any issue here, right? Or did I miss something?

phanen:
You're right. I got confused. We can close it...

PR #890, opened 2024-09-20
Add note about the default hook behaviour

peterjeschke:
I had to look through the code to understand the default value, I think this change will be helpful for new users in the future


ajeetdsouza:
Thank you!

Issue #906, opened 2024-10-05
[Enhancement] Work with environment variables


WiIIiamWei:
I use Windows PowerShell and I tried this (I aliased zoxide to `cd` properly):
```
PS C:\Users\William> cd %temp%
zoxide: no match found
```
Which is expected to be:
```
C:\Users\William\AppData\Local\Temp
```
It would be nice if zoxide could work with these, I believe so called, "enviroment variables".

Could be the similar case of #855.

rajiknows:
we can change the poershell.txt function for __zoxide_cd  
we will check if the path is present in env , if it does than we will set the dir to that path .

ajeetdsouza:
I'm not a Windows user, but isn't `%temp%` the syntax of `CMD.EXE`? I'm not sure I've followed the use case here.

WiIIiamWei:
> I'm not a Windows user, but isn't `%temp%` the syntax of `CMD.EXE`? I'm not sure I've followed the use case here.

Actually yes, blame me for not properly testing this. I only tried this on `cmd` because I don't want to touch my PowerShell profile to disable zoxide, and thought the same syntax would work for PowerShell. However I think it's still a small cool feature if it would work the same in PowerShell.

ajeetdsouza:
I think that would be confusing for many users, I'd suggest wrapping the `z` command in your shell if you'd like this feature.

PR #910, opened 2024-10-07
fixed the issue for powershell env variable

rajiknows:
addressing issue #906 

added the conditional check in powershell.txt

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -27,6 +27,12 @@ function global:__zoxide_pwd {
 
 # cd + custom logic based on the value of _ZO_ECHO.
 function global:__zoxide_cd($dir, $literal) {
+
+    if ($dir.StartsWith('%') -and $dir.EndsWith('%')) {
+        $envVar = $dir.Trim('%')
+        $dir = $env:$envVar
+    }
+    
     $dir = if ($literal) {
         Set-Location -LiteralPath $dir -Passthru -ErrorAction Stop
     } else {


rajiknows:
lint errors can be neglected , it just rearranged the imports due to cargo fmt , which was on by default on my machine .

templates/powershell.txt
```diff
@@ -27,6 +27,12 @@ function global:__zoxide_pwd {
 
 # cd + custom logic based on the value of _ZO_ECHO.
 function global:__zoxide_cd($dir, $literal) {
+
+    if ($dir.StartsWith('%') -and $dir.EndsWith('%')) {
+        $envVar = $dir.Trim('%')
+        $dir = $env:$envVar
```
WiIIiamWei, line 33: I've tested on this, but your code seems to be generating errors when I load the profile on my machine.

I think you meant to write:
```
        $dir = ${env:$envVar}
```

I tried to fix this too and comes up with a different approach (at #911), which might be covering more cases than your solution. Looking for reviews too.

rajiknows: Upon reviewing the repository locally on my machine, I realized the fix could be related to the powershell.txt file, specifically within the __zoxide_cd function, but since I'm currently working on a Linux machine, I wasn't able to test it. I apologize for submitting the PR without proper testing.

ajeetdsouza:
As mentioned in the issue - this is a CMD.EXE feature, not a PowerShell feature.

PR #911, opened 2024-10-07
Fix issues with PowerShell env variables

### LINKED ISSUE #906: [Enhancement] Work with environment variables

WiIIiamWei:
Fixes #906.

`$dir` is expanded using `ExpandEnvironmentStrings` if there is any environment variables.

Tested on my machine (Windows 11, PowerShell 5.1 build 22621) and it works fine.

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -27,6 +27,7 @@ function global:__zoxide_pwd {
 
 # cd + custom logic based on the value of _ZO_ECHO.
 function global:__zoxide_cd($dir, $literal) {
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
As mentioned in the issue - this is a CMD.EXE feature, not a PowerShell feature.

Issue #224, opened 2021-06-06
Smart case search


fenuks:
Hello,
vim has very handy option `smartcase` to do case-sensitive search if there is any capital letter in search query, otherwise search is case-insensitive. I have directory called `V`, and if I type `z V` I am moved, e.g. into `~/.config/nvim` directory. I have to use longer `z parent/V` to give `z` a hint. Therefore, I think it makes sense for `z` to support notion of smart case as well, it would make tool more comfortable to use.

ajeetdsouza:
One question here is -- should `z doc Foo` match `~/Documents/Foo`? I'd think not -- either the whole query should be smartcase or none of it should be.

----

I also think accent normalization (`café -> cafe`) would be great. Possible queries should be:

Accent normalization:
- `z cafe` matches `~/Pictures/Café`
- `z café` does not match `~/Pictures/Cafe`

Case normalization:
- `z cafe` matches `~/Pictures/Cafe`
- `z Cafe` does not match `~/Pictures/cafe`

fenuks:
> One question here is -- should `z doc Foo` match `~/Documents/Foo`? I'd think not -- either the whole query should be smartcase or none of it should be.

I agree, it should be all or nothing. Or perhaps there should be two types of switches, one that applies case-sensitive search globally if there is at least one with capital letter in search query, and other that infers smart case for each word of query individually.

> I also think accent normalization (`café -> cafe`) would be great. Possible queries should be:
> 
> Accent normalization:
> 
>     * `z cafe` matches `~/Pictures/Café`
> 
>     * `z café` does not match `~/Pictures/Cafe`
> 
> 
> Case normalization:
> 
>     * `z cafe` matches `~/Pictures/Cafe`
> 
>     * `z Cafe` does not match `~/Pictures/cafe`

That would be great as well to have, I happen to speak in language with diacritics, but if I were to choose one of the two only, then it would be smart case. ;)

kidonng:
Somewhat related: #114

As for me, I would like zoxide to match case-insensitively all the time, with an option to enable case-sensitive matching. Smart case seems overkill.

ajeetdsouza:
@kidonng I'm curious as to why you'd want to disable smartcase matching. I wouldn't expect anyone to use uppercase in a query unless they were hoping for results with the same uppercase letters in them.

PurpleMyst:
I'm willing to work on this, if that's ok! I'll be trying to send in a PR by tomorrow.

[...]

PR #959, opened 2024-12-22
Add support for case-sensitive queries

rylwin:
Add _ZO_CASE_SENSITIVITY env var to control whether queries should be performed in a case-sensitive manner, keeping the current behavior (case-insensitive) as the default.

Setting _ZO_CASE_SENSITIVITY=case-sensitive changes the query behavior to be case-sensitive.

---

I switched to zoxide a few months ago and have found it to be a much more robust tool than what I was using before. I also enjoy all of the integrations it has.

One thing I miss from my previous jump tool is case-sensitivity: I have some dirs that happen to have capitalization that I can use to get to exactly where I want to go, with a single character, as long as the search is case sensitive.

This PR adds case-sensitivity as an option. This could also lay the groundwork for a "smart case" variant (requested in https://github.com/ajeetdsouza/zoxide/issues/224)

Also, I've done minimal rust, so if anything here is not idiomatic / should be adjusted happy to make any changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- `$_ZO_CASE_SENSITIVITY` to support case-sensitive querying in addition to the
+  default case-insensitive querying.
+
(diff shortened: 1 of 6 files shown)


rylwin:
I've rebased this branch on top of latest main

azaleacolburn (approved):
This looks good to me!

andy5995:
I'd be very interested in this change. I've been using `zoxide` for about 6 weeks now (I heard about it on the Linux Cast youtube channel), and case insensitivity is the only thing I can think that I'd want to change. A very handy utility, thank you.

PR #966, opened 2025-01-05
update nushell init script

Bahex:
1. Made `__zoxide_z` more idiomatic
2. Updated the hook initialization to avoid breakage on nushell's future 0.102.0 release

To expand on hook initialization:
After https://github.com/nushell/nushell/pull/14704#issuecomment-2568660080, starting sub shells would make zoxide's init script throw an error.

With this PR `$env.__zoxide_hooked` is no longer used. Instead, the list of hooks are checked directly to see if zoxide's hook has already been added.

I checked the new init script on both nushell version 0.89.0 and the latest nushell nightly.
Minimum supported version of nushell doesn't need to change and can be kept at 0.89.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,11 +407,13 @@ When calling `zoxide init`, the following flags are available:
   - `--cmd cd` would replace the `cd` command.
 - `--hook <HOOK>`
   - Changes how often zoxide increments a directory's score:
+
     | Hook            | Description                       |
     | --------------- | --------------------------------- |
     | `none`          | Never                             |
     | `prompt`        | At every shell prompt             |
     | `pwd` (default) | Whenever the directory is changed |
+
 - `--no-cmd`
   - Prevents zoxide from defining the `z` and `zi` commands.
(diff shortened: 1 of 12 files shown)


KamilKleina:
Is it ionly my problem that zoxide stopped adding paths to it's db?

Alan-TheGentleman:
@KamilKleina not at all, the only way it works is using `zoxide add /path` but is tedious

Alan-TheGentleman:
@KamilKleina  FIXED !!

this is my pr for my personal configs which solve the issue for now until a proper fix !!
https://github.com/Gentleman-Programming/Gentleman.Dots/commit/c00622d8f18eec103376a9ba2556d6a06da94df1

zoxide adds new paths and everything. Enjoy !

ajeetdsouza:
Thanks! Loving the new Nushell syntax :)

KamilKleina:
Guys,
I think it still doesn't work. I updated my zoxide.nu with new template and it still doesn't add dirs to zoxide

PR #980, opened 2025-02-08
Stream filter logic

SeeStarz:
Proof of concept feature improvement related to #260 

This adds a new option to allow exact matching of sub path (the ones in between slashes). e.g. `zoxide query rust` matches `rust` but not `rust-analyzer`. However I haven't been able to understand how to integrate this option into CLI options, as such this pull request isn't complete yet.

I personally think just being able to exact match directories would help with QoL especially since the ones most vulnerable to regex matching is short names anyways.

Any feedback would be welcome!

### DIFF
--- a/src/db/stream.rs
+++ b/src/db/stream.rs
@@ -1,5 +1,7 @@
+use std::ffi::OsString;
 use std::iter::Rev;
 use std::ops::Range;
+use std::path::{Component, PathBuf};
 use std::{fs, path};
 
 use glob::Pattern;
@@ -37,6 +39,10 @@ impl<'a> Stream<'a> {
                 if dir.last_accessed < self.options.ttl {
                     self.db.swap_remove(idx);
                 }
+                continue;
+            }
+
+            if !self.filter_by_exact(&dir.path) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


SeeStarz:
Now that it comes to shell integration I don't think I am qualified to progress any further, I do hope this will be somewhat useful if anyone has similar ideas.

The added feature is `-e` or `--exact` flag on `zoxide query`

dev-ardi:
Your git history is whack since you merged instead of rebasing

SeeStarz:
I completely missed that, thanks! 
I'm not sure where I f'd up but nothing a good old force push can't solve.
Note: If anyone has seen my PR before, both commits are merged into one forming a complete feature (albeit missing shell integration)

SeeStarz:
Hi it's been a while, average random guy making a PR am i right. Also thanks for reading my PR and reviewing it, and sorry it took so long to reply.

Anyways the best feature explanation i came up with while being consistent with others is `/// Only match path components exactly (e.g. bin != sbin)` It's really hard to convey that `bin` matches `/usr/bin` and `/bin` but not `/sbin` or similar.

Upon reconsideration, it's apparent that my implementation doesn't play nicely with any queries with `/` in it. Meanwhile testcases suggests that slashes are part of the feature. So it seems like if the intention is to match `/usr/bin` but not `/usr/sbin` one solution is to simply match for `/bin`. However the problem shifts to filtering out `/usr/binn` or any other forms of suffix.

For this reason i'm changing this PR back to draft.

My personal interpretation concludes that  the easiest way to do so is modifying `filter_by_keywords` directly to account for the `--exact` option, but I'm not sure if I have the time to understand how this function works in full picture to not mess things up.

PR #1001, opened 2025-02-25
Fixes #995: Wildcard expansion for PowerShell

### LINKED ISSUE #995: PowerShell: cd dir* Pattern Matching Doesn't Work Due to -LiteralPath
#### Description  
In PowerShell, `cd desk*` does not work as expected when using **zoxide**. Normally, PowerShell supports wildcard expansion in paths, but `zoxide` restricts this behavior. Instead of navigating to a matching directory (e.g., `Desktop`), it fails or does not expand the pattern.  

> [!IMPORTANT]
> Upon reviewing the implementation, the issue seems to be caused by the use of the `-LiteralPath` parameter in the `__zoxide_cd` function. `-LiteralPath` prevents wildcard expansion, which explains why commands like `cd desk*` do not work.  

#### Expected Behavior  
PowerShell’s `cd desk*` should work similarly to its native behavior, where it matches a directory using wildcard expansion (e.g., resolving `desk*` to `Desktop`).  

#### Actual Behavior  
The command does not work as expected because `zoxide` uses `Set-Location -LiteralPath`, which prevents wildcard expansion.  

#### Suspected Cause  
The issue appears to be due to the `__zoxide_cd` function in the PowerShell integration:  

```powershell
# __zoxide_cd :
param($dir, $literal)
$dir = if ($literal) {
    Set-Location -LiteralPath $dir -Passthru -ErrorAction Stop
} else {
   ...
}
```

Since `-LiteralPath` is used when `$literal` is true, it bypasses PowerShell's wildcard handling.  

#### Suggested Fix  
Consider modifying the function to allow wildcard expansion when applicable. Instead of `-LiteralPath`, using `-Path` should enable this functionality.   
Or modify the function `__zoxide_z`, conditions to call `__zoxide_cd` with `$literal = $false` if the input contains wildcards.

```powershell
# __zoxide_z :
...
 elseif ($args.Length -eq 1 -and ($args[0] -eq '-' -or $args[0] -eq '+')) {
        __zoxide_cd $args[0] $false
    }
    elseif ($args.Length -eq 1 -and (Test-Path $args[0] -PathType Container)) {
        __zoxide_cd $args[0] $true
    }
...
```

#### Environment  
- **OS:** Windows 11  
- **PowerShell Version:** 7.5.0  
- **zoxide Version:** 0.9.7 

#### Additional Notes  
- This issue seems specific to PowerShell, as wildcard matching works differently in other shells (e.g., Bash).  
- If `-LiteralPath` is required for other use cases, an alternative could be to conditionally use `-Path` when the input contains wildcards.

benzaria:
Modifying the `__zoxide_z` function to support wildcard expansion in the current directory, allowing users to navigate to matching directories dynamically. This update enhances usability by resolving wildcard expressions (*, ?) before attempting to change directories, ensuring a smoother experience when working with complex directory structures

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -107,7 +107,7 @@ function global:__zoxide_z {
         __zoxide_cd $args[0] $false
     }
     elseif ($args.Length -eq 1 -and (Test-Path $args[0] -PathType Container)) {
-        __zoxide_cd $args[0] $true
+        __zoxide_cd $args[0] $(Test-Path -LiteralPath $args[0])
     }
     else {
         $result = __zoxide_pwd
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -106,8 +106,11 @@ function global:__zoxide_z {
     elseif ($args.Length -eq 1 -and ($args[0] -eq '-' -or $args[0] -eq '+')) {
         __zoxide_cd $args[0] $false
     }
-    elseif ($args.Length -eq 1 -and (Test-Path $args[0] -PathType Container)) {
-        __zoxide_cd $args[0] $(Test-Path -LiteralPath $args[0])
+    elseif ($args.Length -eq 1 -and (Test-Path $args[0] -PathType Container -LiteralPath)) {
+        __zoxide_cd $args[0] $true
+    }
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Thanks so much for the fix!

PR #1009, opened 2025-03-05
Explain that `_ZO_EXCLUDE_DIRS` doesn't remove existing entries.

SamirTalwar:
When trying to use `_ZO_EXCLUDE_DIRS`, I was confused why it wasn't working with `zi`. It turns out that it worked well, but some of the directories I wanted to exclude were already in the database.

I have added a line to the README to make this clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,6 +437,8 @@ Environment variables[^2] can be used for configuration. They must be set before
     | Windows             | `;`       | `$HOME;$HOME/private/*` |
 
   - By default, this is set to `"$HOME"`.
+  - This does not exclude directories that have already been added to the database.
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
`zi` should not show you any directories that are in `_ZO_EXCLUDE_DIRS`. Are you sure you've excluded them correctly?

SamirTalwar:
Yes, because after excluding them _and_ removing them from the DB, they didn't show up again.

SamirTalwar:
Tested again and it seems to work even without touching the database. I have no idea what happened the first time around (and I did double-check).

Apologies for the waste of time.

Issue #952, opened 2024-12-07
`zoxide: no match found` can be too easily confused with `zoxide: command not found`


AltiorRonan:
When you type a command like `z an-absoloutly-unkown-directory`, the response is `zoxide: no match found` which can be too easily confused with the fact that the command `z` does not exist (`zoxide: command not found`).

A better message would be: `zoxide: no match found for 'an-absoloutly-unkown-directory'` which is a better distinction between zoxide vs OS error message.

azaleacolburn:
Generally, I agree. One issue with this though is that multiple `keywords` can be used in a query, e.g., `zoxide foo bar`.

I propose that it be formatted as such:
`zoxide: no match found for 'foo' and 'bar'

ajeetdsouza:
That seems to me like it would be very verbose. It's also redundant, since the query on most shells would be visible right above the error:

```console
$ z foo bar
zoxide: no match found for 'foo bar'
```

It would also be slightly confusing in places:

```console
$ z "foo bar" baz
zoxide: no match found for '"foo bar" baz'
```

Lastly, the case we're optimizing for (`zoxide: command not found`) seems like it would not happen very often on most setups - it seems like we're optimizing for 1% of situations at the expense of the remaining 99%.

martijnn:
Stumbled upon this. Was trying to figure out the correct installation and configuration in wsl, but apparently I suddenly went from command not found to no match found. Somehow, I didn't pick up that change.

PR #1016, opened 2025-03-15
Clearer error messages when no matches are found

azaleacolburn:
## Description
Adds clearer error messages when no matches are found in a query as per #952.

An error message for the command:
`./target/debug/zoxide query foo bar baz`
Would now read:
`zoxide: no match found for 'foo', 'bar', and 'baz'`
Instead of:
`zoxide: no match found`

The intention is to differentiate the error for `zoxide: command not found` and to provide clarity.

> [!WARNING]
> This PR adds the method `format_keywords` to the `Query` struct. It is run during every query because of how expected errors are managed with context. It clones `self.keywords`; if this is undesirable, it can be refactored.

### DIFF
--- a/src/cmd/query.rs
+++ b/src/cmd/query.rs
@@ -126,18 +126,14 @@ impl Query {
     /// ## WARNING
     /// Clones `self.keywords`
     fn format_keywords(&self) -> String {
-        if self.keywords.is_empty() {
-            return String::new();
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
I'm not convinced that this feature would be useful, see: https://github.com/ajeetdsouza/zoxide/issues/952#issuecomment-2749354656

Issue #1014, opened 2025-03-11
Feature: Add bookmark


OshinoShinobu-Chan:
Add a subcommand like `zoxide bookmark foo ~/foo/src`, so you can use `z foo` to jump to `~/foo/src`. This may solve the problem, where multiple projects all have folder `src`. You can add different bookmarks for different `src`s in different projects, and can jump between them conveniently.

azaleacolburn:
I support this and would be happy to open a PR.

azaleacolburn:
I'm not particularly familiar with this project, so I might not be understanding the nuances of the database, but I believe this could be accomplished by adding a `bookmarks` field to the `Database` struct. Then, the bookmark command would add to both the `dirs` and the `bookmarks` field. The query command would first check the database for exact bookmark matches before matching on `dirs`.

lengyijun:
I don't like bookmark system.
The database of zoxide should be kept simple.
Bookmark should be done in other tools rather than zoxide.

azaleacolburn:
How would these other tools integrate with zoxide? Or do you mean a completely separate tool for bookmarking directories? Currently the only dababase change is making it a tuple, which doesn't actually add meaningful complexity in my opinion.

lengyijun:
A completely separate tool for bookmarking directories

lengyijun:
`z foo` choose bookmark first: I really doubt this design

lengyijun:
I guess supporting bookmark requires 1000 loc
- completion for bookmark
- fuzzy select on bookmark (both key and value)

lengyijun:
I hope intuitive prs to be merged before: #1027 #1039

azaleacolburn:
>     * completion for bookmark
> 
>     * fuzzy select on bookmark (both key and value)

I agree this would become problematic. Could we have the bookmark feature without these additions? What do you think @ajeetdsouza?

AlexandreDecan:
> I don't like bookmark system. The database of zoxide should be kept simple. Bookmark could be done in other tools rather than zoxide.

To what extent will a bookmark system affect the simplicity of the database? Is there any side-effect for users that won't use or do not prefer to use this bookmark system?

lengyijun:
https://github.com/olafurw/qwe
Directory bookmarker

PR #1019, opened 2025-03-19
New Bookmark Command

azaleacolburn:
## Description
Adds a new subcommand called `bookmark` as per #1014

## Usage
`zoxide bookmark foo ~/bar` creates a bookmark in the database called `foo` which aliases to `bar`.

Then, `zoxide query foo` automatically outputs `~/bar` before even opening a `Stream`. However, it doesn't match, so `zoxide query fooo` ignores the bookmarking system and matches `dirs` on `fooo`.

This PR restructures the database, by turning `dirs` into a tuple containing the old `dirs` (a vector of directories) and a bookmarks hashmap. They must be under the same field so they can be serialized and deserialized together, while obeying the `self_referencing` struct system.

> [!WARNING]
> Because of this restructuring, old databases will no longer be deserializable. 

However, this PR also adds additional code to the deserialize command, which will fall back on deserializing just the vector of `Dirs`, then reserializing the database with those dirs and an empty hashmap.
So existing zoxide instances should still operate perfectly, but old databases will be changed.


azaleacolburn:
> I don't like bookmark system. Bookmark should be done in other tools rather than zoxide.

Discussions about the viability of bookmarking as a feature can be had in #1014.

Issue #997, opened 2025-02-23
FEATURE REQUEST: Ability for zi to go directly to single result


jasonlyle88:
If `zi` comes up with only 1 result, it would be nice if we were able to configure it so that it could automatically go directly to that result! Maybe an environment variable that could be set?

jasonlyle88:
That's why I said it should be a configurable option. I do t want to go changing the default behavior that everyone has come to expect from zi. So by default, zi would still behave the same. But if we set this new environment variable (_ZO_ZI_GO_TO_ONLY_RESULT=true or something like that?), then this new behavior would be enabled.

When I originally picked up zoxide, I had considered aliasing cd to zi... But quickly decided against it when I discovered that there was no way to make zi go to a single result.

This feature makes that more of a possibility. Maybe still not a great option, but it makes it more realistic at least!

azaleacolburn:
The PR #1026 will creates an environment variable that allows additional options to be passed into `fzf` when running `zi`. Once that PR is merged, you'll be able to set the behavior you've detailed with:
```bash
export _ZO_FZF_EXTRA_OPTS="--select-1"
```

> [!NOTE]
> The exact environment variable name might change, so be sure to check the docs after the PR is merged.

PR #1022, opened 2025-03-21
zi automatically selects only initial result

azaleacolburn:
## Description
`zi foo` will now automatically select the only initial result if applicable, as per #997.

## Usage
The command `zoxide query --interactive foo` will now automatically select the only result initially returned, if there is only one. If there is initially more than one result, but further fuzzy-finding narrows it down to just one, that final result will not be automatically selected.

## Environment Variable
A new environment variable `_ZO_ZI_ONLY_RESULT` has been introduced which toggles the behavior described above with the current behavior of `zi`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,7 +453,8 @@ Environment variables[^2] can be used for configuration. They must be set before
     database.
 - `_ZO_ZI_ONLY_RESULT`
   - When set to 1, `zi foo` will automatically select the only result `fzf` returns.
-  - However, if only one result is available during selection in the `fzf` menu, it will not be automatically selected.
+  - However, if only one result is available during selection in the `fzf` menu,
+    it will not be automatically selected.
 
 ## Third-party integrations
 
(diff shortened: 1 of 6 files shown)


ajeetdsouza:
Hey @azaleacolburn, thanks for sending in these PRs. I really appreciate the help! One small request though - can you tag me before picking up something? I may have a different implementation in mind, so I could probably save you some time there.

For this particular issue, I think this would be the cleanest solution:

- Currently, zoxide uses certain default flags for `fzf`, which the user can override with `_ZO_FZF_OPTS`.
- We can add a flag called `_ZO_FZF_OPTS_EXTRA` that _adds_ flags on top of the default ones, rather than replacing them.
- Once that works, users can get this behaviour with the following setup:

```zsh
export _ZO_FZF_OPTS_EXTRA="--select-1"
```

ValJed:
@ajeetdsouza @azaleacolburn Just seing this PR and your answer, funny but I just opened a PR to support `_ZO_FZF_EXTRA_OPTS`.
https://github.com/ajeetdsouza/zoxide/pull/1026

I can rename the flag if you prefer.

azaleacolburn:
@ajeetdsouza Thank you! I'll be sure to tag you going forward. I didn't realize the `--select-1` option existed. I agree that it's clearly the cleanest solution. I'll close this PR and respond to #997 letting them know about #1026 and the `select-1` flag.

Issue #618, opened 2023-09-08
Allow `_ZO_FZF_OPTS` to not override the useful default values


clenemt:
`_ZO_FZF_OPTS` override all the values at the moment. Would it be possible to have another option that keeps the default values?
Maybe `_ZO_FZF_ADDITIONAL_OPTS` or something similar.
Happy to push a PR if needed.

stablestud:
+1

Yes please! I would've liked to remove the FZF borders with `--no-border`, to align it with the FZF shell completion,  but setting the FZF opts removes all Zoxide defaults.

> Maybe `_ZO_FZF_ADDITIONAL_OPTS` or something similar.

Instead of `_ZO_FZF_ADDITIONAL_OPTS` I would vote for `_ZO_FZF_EXTRA_OPTS`

ValJed:
I'm digging up this issue, just created a PR for this:
https://github.com/ajeetdsouza/zoxide/pull/1026

juliogc:
I've solved the issue with this solution

```bash
export _ZO_FZF_OPTS="
  ${FZF_DEFAULT_OPTS}
  # extra opts
"
```

ElgebarOri:
🙌🙌🙌I raised my hands in approval！🙌🙌🙌

ValJed:
Yes! I still would find cleaner to have an extra env variable.
I would like to support it in yazi afterward that uses fzf and zoxide, but it creates `FZF_DEFAULT_OPTS` programmatically so it wouldn't be that easy in this case.

PR #1026, opened 2025-03-24
feat: support fzf extra options

ValJed:
## Description

Add support for a new env variable `_ZO_FZF_EXTRA_OPTS` to allow adding flags to the default ones, instead of erasing them.

This has been asked in [this issue](https://github.com/ajeetdsouza/zoxide/issues/618).

## Usage

Simply set `_ZO_FZF_EXTRA_OPTS` instead of `_ZO_FZF_OPTS` to append flags to the default ones.

### DIFF
--- a/src/cmd/query.rs
+++ b/src/cmd/query.rs
@@ -3,7 +3,7 @@ use std::io::{self, Write};
 use anyhow::{Context, Result};
 
 use crate::cmd::{Query, Run};
-use crate::config::{self};
+use crate::config;
 use crate::db::{Database, Epoch, Stream, StreamOptions};
 use crate::error::BrokenPipeHandler;
 use crate::util::{self, Fzf, FzfChild};
--- a/src/cmd/query.rs
+++ b/src/cmd/query.rs
@@ -3,7 +3,7 @@ use std::io::{self, Write};
 use anyhow::{Context, Result};
 
 use crate::cmd::{Query, Run};
-use crate::config;
+use crate::config::{self};
 use crate::db::{Database, Epoch, Stream, StreamOptions};
 use crate::error::BrokenPipeHandler;
 use crate::util::{self, Fzf, FzfChild};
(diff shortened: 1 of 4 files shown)


src/cmd/query.rs
```diff
@@ -91,29 +91,29 @@ impl Query {
...
-                "--info=inline",
-                "--layout=reverse",
-                // Display
-                "--tabstop=1",
-                // Scripting
-                "--exit-0",
-            ])
-            .enable_preview()
+            let default_args = config::fzf_default_args();
+            let args = if let Some(fzf_extra_opts) = config::fzf_extra_opts() {
```
azaleacolburn, line 103: Since this returns values for both branches, it should probably be a `match` expression. `if let` expressions are generally used when there's some code that ought to be conditionally executed, rather than being assigned to a variable in the parent scope.

src/cmd/query.rs
```diff
@@ -91,29 +91,29 @@ impl Query {
 
     fn get_fzf() -> Result<FzfChild> {
         let mut fzf = Fzf::new()?;
-        if let Some(fzf_opts) = config::fzf_opts() {
+        if let Some(mut fzf_opts) = config::fzf_opts() {
```
azaleacolburn, line 94: This could be a match expression

src/cmd/query.rs
```diff
@@ -91,29 +91,29 @@ impl Query {
...
+            if let Some(fzf_extra_opts) = config::fzf_extra_opts() {
```
azaleacolburn, line 95: This is the right way to use an `if let`, since there's no other execution path.

azaleacolburn (changes requested):

azaleacolburn:
```patch
diff --git a/src/cmd/query.rs b/src/cmd/query.rs
index adaeff7..652e624 100644
--- a/src/cmd/query.rs
+++ b/src/cmd/query.rs
@@ -3,7 +3,7 @@ use std::io::{self, Write};
 use anyhow::{Context, Result};
 
 use crate::cmd::{Query, Run};
-use crate::config;
+use crate::config::{self};
 use crate::db::{Database, Epoch, Stream, StreamOptions};
 use crate::error::BrokenPipeHandler;
 use crate::util::{self, Fzf, FzfChild};
@@ -91,29 +91,33 @@ impl Query {
 
     fn get_fzf() -> Result<FzfChild> {
         let mut fzf = Fzf::new()?;
-        if let Some(mut fzf_opts) = config::fzf_opts() {
-            if let Some(fzf_extra_opts) = config::fzf_extra_opts() {
-                fzf_opts.push(" ");
-                fzf_opts.push(fzf_extra_opts);
+        match config::fzf_opts() {
+            Some(mut fzf_opts) => {
+                if let Some(fzf_extra_opts) = config::fzf_extra_opts() {
+                    fzf_opts.push(" ");
+                    fzf_opts.push(fzf_extra_opts);
+                }
+
+                fzf.env("FZF_DEFAULT_OPTS", fzf_opts)
             }
+            None => {
+                let default_args = config::fzf_default_args();
+                let args = match config::fzf_extra_opts() {
+                    Some(fzf_extra_opts) => {
+                        let extra_fzf_args_list: Vec<String> = fzf_extra_opts
+                            .to_str()
+                            .unwrap_or_default()
+                            .split_whitespace()
+                            .map(|arg| String::from(arg))
+                            .collect();
+
+                        vec![default_args, extra_fzf_args_list].concat()
+                    }
+                    None => default_args,
+                };
 
-            fzf.env("FZF_DEFAULT_OPTS", fzf_opts)
-        } else {
-            let default_args = config::fzf_default_args();
-            let args = if let Some(fzf_extra_opts) = config::fzf_extra_opts() {
-                let extra_fzf_args_list: Vec<String> = fzf_extra_opts
-                    .to_str()
-                    .unwrap_or_default()
-                    .split_whitespace()
-                    .map(|arg| String::from(arg))
-                    .collect();
-
-                vec![default_args, extra_fzf_args_list].concat()
-            } else {
-                default_args
-            };
-
-            fzf.args(args).enable_preview()
+                fzf.args(args).enable_preview()
+            }
         }
         .spawn()
     }
```

src/cmd/query.rs
ValJed: That's right, updated.

src/cmd/query.rs
ValJed: Indeed, udpated too!

azaleacolburn (approved):

azaleacolburn (approved):

ValJed:
@ajeetdsouza Would love to see this merged 🙏🏼 
Thanks in advance!

azaleacolburn:
@ajeetdsouza Several issues would be resolved by this feature, which you've already endorsed. I would love to get your review on this PR.

j8s0n:
Will this feature make it into zoxide? When I run `z foo` and it goes to the wrong directory, I have to run a second command `zi foo` and select the other one. I don't want to run `zi foo` every time, because most of the time there's only one target.

jasonlyle88:
@azaleacolburn @ajeetdsouza is there a problem with this PR it's been waiting for a review for almost a year now. Does something need to be done to move it along?

azaleacolburn:
Hi! I'm not a maintainer on this repository, only @ajeetdsouza has permissions to merge this PR. I currently see no problems with nor reasons to not merge this PR.

Issue #944, opened 2024-12-02
[query] please support a --basedir parameter to filter roots


jessefarinacci:
I often want to run `zoxide query --interactive` in a large project and have the results all be at the project root where I'm invoking `zoxide`, not just the ones zoxide thinks are the most popular.

I propose a new `--basedir [path]` parameter to the `query` subcommand. If user passes `--all` we can error out or just preference one or the other. With `--list` and `--interactive` it will filter all results for having the same fully qualified path prefix specified by `--basedir` parameter, with the `$CWD` being the default value.

The query subcommand already supports an `--exclusion` mechanism for filtering. This new option to effectively invert that logic will also be valuable. Thanks!

jessefarinacci:
Hi, I'm proposing a feature of `--basedir [optional specified path parameter, or $cwd by default]` argument to allow users to filter `zoxide` suggestions that start with the same root path. Maybe some examples will be helpful, imagine a server log directory broken down roughly by timestamp part directories:

```
❯ zoxide query --list
/tmp/zoxide/2025
/tmp/zoxide/2025/02
/tmp/zoxide/2025/03
/tmp/zoxide
/tmp/zoxide/2025/02/05
/tmp/zoxide/2025/02/04
/tmp/zoxide/2025/03/01
/tmp/zoxide/2025/03/09
/tmp/zoxide/2025/03/17
/tmp/zoxide/2025/02/14
/tmp/zoxide/2025/02/02
/tmp/zoxide/2025/01/06
/tmp/zoxide/2025/01/05
/tmp/zoxide/2025/01/02
/tmp/zoxide/2025/01
/tmp/zoxide/2025/01/01
.......
```

```
❯ zoxide query --list --basedir /tmp/zoxide/2025/01
/tmp/zoxide/2025/01/06
/tmp/zoxide/2025/01/05
/tmp/zoxide/2025/01/04
/tmp/zoxide/2025/01/03
/tmp/zoxide/2025/01/02
/tmp/zoxide/2025/01
/tmp/zoxide/2025/01/01
........
```

```
❯ cd /tmp/zoxide/2025/02
❯ zoxide query --list --basedir
/tmp/zoxide/2025/02
/tmp/zoxide/2025/02/05
/tmp/zoxide/2025/02/04
/tmp/zoxide/2025/02/14
/tmp/zoxide/2025/02/09
/tmp/zoxide/2025/02/08
/tmp/zoxide/2025/02/06
[...]

PR #1027, opened 2025-03-25
--basedir query option

azaleacolburn:
## Description
Adds the `--basedir foo` flag to `zoxide query` as per #944. It restricts results to the given directory and is thus incompatible with `--exclude bar` as they could be passed the same argument.


lengyijun:
Idea:  add a shell function `zz` : search subdirs: `--basedir=(pwd)`

aldanor:
Looks pretty useful, especialy being able to pass pwd in. Any chances this could be merged? 🤔

ajeetdsouza:
Merged, thanks!

PR #1030, opened 2025-03-29
Add `--score` option

### LINKED ISSUE #1010: Allow adding a directory without increasing its score/rank, e.g. `zoxide add --no-increment <paths>`
I use zoxide on multiple machines, and I really want two things:

1. Anywhere I have been, sorted by frecency → zoxide does a great job of this
2. The core set of directories I know I might need to go to are somewhere in the list → this is where this feature would be useful.

For `1.` having the hook on changing directory to update the frecency works really well. But if I haven't been to that directory on this machine before, it won't show up at all.

So I have a script that runs as part of my updates that basically does this:

```zsh
zoxide_dirs=(
  /opt/homebrew/Library/Taps/*/* # Brew taps.
  ~/.local/share/nvim/lazy/*/    # Neovim plugins.
  ~/.local/share/zsh/plugins/*/  # Zsh plugins.
)

# Zoxide should ignore inputs that don't exist, so put lots of things here.
zoxide add "${zoxide_dirs[@]}"
```

This works great for making sure that I have the stuff I occasionally use in the list somewhere. However this also means I'm increasing the rank of each directory as I add it. I know I could write a custom function that looks at each dir in turn, and checks whether it's in the `zoxide query --list` output, but that's a bunch of looping and parsing.

Instead it would be great if there was a way to say "add this dir if it's not present, but if it is, don't increase its rank or score.

azaleacolburn:
## Description
Adds the `--score` flag to the `zoxide add` command. It dictates the initial rank given to the new entry, or the value to increment it's existing rank by if the entry already exists.

> [!NOTE]
> While negative `score` values are allow (for decrementing rank), the minimum possible rank of an entry is still `0.0`.

Resolves #1010 when an initial score of zero is passed in such as `zoxide add --score 0.0 foo`.

### DIFF
--- a/contrib/completions/_zoxide
+++ b/contrib/completions/_zoxide
@@ -30,6 +30,8 @@ _zoxide() {
         case $line[1] in
             (add)
 _arguments "${_arguments_options[@]}" : \
+'-s+[The rank to increment the entry if it exists or initialize it with if it doesn'\''t]:SCORE: ' \
+'--score=[The rank to increment the entry if it exists or initialize it with if it doesn'\''t]:SCORE: ' \
 '-h[Print help]' \
 '--help[Print help]' \
 '-V[Print version]' \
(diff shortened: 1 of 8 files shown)


ajeetdsouza:
Thanks, this is great!

PR #1036, opened 2025-04-08
Don't downcast PROMPT_COMMAND

SuperSandro2000:
Since bash 5.0 PROMPT_COMMAND is an array. Lets not turn it into a string. Copied from https://github.com/direnv/direnv/blob/master/internal/cmd/shell_bash.go#L19-L23


azaleacolburn (approved):

SuperSandro2000:
@ajeetdsouza can you take a look?

ajeetdsouza:
Thanks!

PR #1038, opened 2025-04-10
powershell: Ignore errors from `__zoxide_hooked`

brian6932:
This line currently logs to `$error` every time for no reason, it'll still work in strict mode this way.


brian6932:
Ah gotcha, easy solution.

ajeetdsouza:
Looks great, thank you!

PR #1042, opened 2025-04-15
Add `z +` to docs

asasine:
I noticed `z +` was supported syntax by the  completions for my shell so I added it to the tutorial and man page.


azaleacolburn (commented):
I don't think `z +` is supported, if you think it should be you can open an issue requesting it.

ajeetdsouza:
That's correct - `z +` is a shell feature that's not available on all shells. If someday we make it a part of all shells, we can add it to the docs.

PR #1048, opened 2025-05-03
Add fixes to Bash integration

### LINKED ISSUE #1043: cd foo<SPACE><TAB> shows fzf, but does not change into selected dir on return
### Description

I am using zoxide 0.9.7 and it is initialized like this:

```
eval "$(zoxide init --cmd cd bash)"
```

### `cd foo<SPACE><TAB>`

When I use `cd foo<SPACE><TAB>`, the options are shown with fzf. However, when I select one and hit enter, I get my prompt replaced with: `cd foo z#/path/here/`

https://github.com/user-attachments/assets/b241be5c-3af2-4040-9dc6-60f1a7e49ba4

### `cdi foo<ENTER>`

But when I do a `cdi foo<ENTER>`, fzf shows up and upon return I am changed into the selected directory.

https://github.com/user-attachments/assets/a90169b4-6587-4273-b348-f708fad6279f

akinomyoga:
This PR includes four fixes, each corresponding to the respective commit. Please check the commit messages; the details are described in the commit messages.

Fixes #1043.

### DIFF
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -86,6 +86,8 @@ function __zoxide_doctor() {
 # When using zoxide with --no-cmd, alias these internal functions as desired.
 #
 
+__zoxide_z_prefix='z#'
+
 # Jump to a directory using only keywords.
 function __zoxide_z() {
     __zoxide_doctor
@@ -99,6 +101,10 @@ function __zoxide_z() {
         __zoxide_cd "$1"
     elif [[ $# -eq 2 && $1 == '--' ]]; then
         __zoxide_cd "$2"
+    elif [[ ${@: -1} == "${__zoxide_z_prefix}"?* ]]; then
+        # shellcheck disable=SC2124
+        \builtin local result="${@: -1}"
+        __zoxide_cd "{{ "${result:${#__zoxide_z_prefix}}" }}"
     else
         \builtin local result
         # shellcheck disable=SC2312
@@ -157,6 +163,11 @@ if [[ ${BASH_VERSINFO[0]:-0} -eq 4 && ${BASH_VERSINFO[1]:-0} -ge 4 || ${BASH_VER
         elif [[ -z ${COMP_WORDS[-1]} ]]; then
             # shellcheck disable=SC2312
             __zoxide_result="$(\command zoxide query --exclude "$(__zoxide_pwd)" --interactive -- "{{ "${COMP_WORDS[@]:1:${#COMP_WORDS[@]}-2}" }}")" && {
+                # In case the terminal does not respond to \e[5n or another
+                # mechanism steals the response, it is still worth completing
+                # the directory in the command line.
+                COMPREPLY=("${__zoxide_z_prefix}${__zoxide_result}/")
+
                 \builtin bind -x '"\e[0n": __zoxide_z_complete_helper'
                 \builtin printf '\e[5n'
             }
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -168,8 +168,11 @@ if [[ ${BASH_VERSINFO[0]:-0} -eq 4 && ${BASH_VERSINFO[1]:-0} -ge 4 || ${BASH_VER
                 # the directory in the command line.
                 COMPREPLY=("${__zoxide_z_prefix}${__zoxide_result}/")
 
-                \builtin bind -x '"\e[0n": __zoxide_z_complete_helper'
-                \builtin printf '\e[5n'
+                # Note: We here call "bind" without prefixing "\builtin" to be
+                # compatible with frameworks like ble.sh, which emulates Bash's
+                # builtin "bind".
+                bind -x '"\e[0n": __zoxide_z_complete_helper'
+                \builtin printf '\e[5n' >/dev/tty
             }
         fi
     }
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -157,9 +157,6 @@ if [[ ${BASH_VERSINFO[0]:-0} -eq 4 && ${BASH_VERSINFO[1]:-0} -ge 4 || ${BASH_VER
         elif [[ -z ${COMP_WORDS[-1]} ]]; then
             # shellcheck disable=SC2312
             __zoxide_result="$(\command zoxide query --exclude "$(__zoxide_pwd)" --interactive -- "{{ "${COMP_WORDS[@]:1:${#COMP_WORDS[@]}-2}" }}")" && {
-                \builtin bind '"\e[0n": redraw-current-line'
-                \builtin printf '\e[5n'
-
                 # shellcheck disable=SC2016
                 \builtin bind -x '"\e[0n": __zoxide_z_complete_helper "${result}"'
                 \builtin printf '\e[5n'
--- a/templates/bash.txt
+++ b/templates/bash.txt
@@ -157,8 +157,7 @@ if [[ ${BASH_VERSINFO[0]:-0} -eq 4 && ${BASH_VERSINFO[1]:-0} -ge 4 || ${BASH_VER
         elif [[ -z ${COMP_WORDS[-1]} ]]; then
             # shellcheck disable=SC2312
             __zoxide_result="$(\command zoxide query --exclude "$(__zoxide_pwd)" --interactive -- "{{ "${COMP_WORDS[@]:1:${#COMP_WORDS[@]}-2}" }}")" && {
-                # shellcheck disable=SC2016
-                \builtin bind -x '"\e[0n": __zoxide_z_complete_helper "${result}"'
+                \builtin bind -x '"\e[0n": __zoxide_z_complete_helper'
                 \builtin printf '\e[5n'
             }
         fi


akinomyoga:
[The CI failure](https://github.com/ajeetdsouza/zoxide/actions/runs/14811054355/job/41585752787) seems unrelated to the changes in this PR. [The same error](https://github.com/ajeetdsouza/zoxide/actions/runs/14430060292/job/40463769318) is issued against the current `main` branch.

ajeetdsouza:
Hey, thanks for this! Would it make sense to just use the regular `bind` instead of `\builtin bind` so that it works the same way with `ble.sh`?

akinomyoga:
> Hey, thanks for this! Would it make sense to just use the regular `bind` instead of `\builtin bind`

Thank you! If you can accept calling `bind` without `\builtin`, it would be more stable with `ble.sh`. I pushed another commit 1c9170f.

> so that it works the same way with `ble.sh`?

To make it work the same way with `ble.sh`, `printf '\e[0n'` also needs to be adjusted; the stdout needs to be explicitly redirected to `/dev/tty`. I also added the necessary redirection in commit 1c9170f. In addition, the user needs to add the following `ble.sh` setting (because `ble.sh` discards unrecognized CSI sequences with the default settings):

```bash
# In ~/.blerc or ~/.config/blesh/init.sh

bleopt decode_error_cseq_discard=
```

tessus:
Thanks so much for the fix @akinomyoga and thank you a bunch as well @ajeetdsouza for your willingness to make it work with ble!

akinomyoga:
> In addition, the user needs to add the following `ble.sh` setting (because `ble.sh` discards unrecognized CSI sequences with the default settings):
> 
> ```shell
> # In ~/.blerc or ~/.config/blesh/init.sh
> 
> bleopt decode_error_cseq_discard=
> ```

In `ble.sh`, I added a new key definition for `\e[0n` with the name <kbd>dsr0</kbd> so that the user does not need to write the above setting (commit https://github.com/akinomyoga/ble.sh/commit/adf53ed356304a91634e3f690e05c24818f0e042). I also adjusted the commit message of the last commit of this PR.

akinomyoga:
I fixed the shellcheck errors, rebased, and pushed.

I tried to run `shellcheck --enable=all`, but I still could not reproduce the errors that the shellcheck on CI produces. Instead, I modified the respective commits based on the error messages on the CI and force-pushed. However, the CI does not seem to start again.

ajeetdsouza:
This is really cool, thank you so much!

The reason you're not seeing the shellcheck error locally is likely because you're running a different version of shellcheck. That should be fine, I can fix that - but could you explain how the `__zoxide_original_keybinding` function works? I'm wondering if there's a way to do it with just shell builtins (i.e. no `awk`).

akinomyoga:
> could you explain how the `__zoxide_original_keybinding` function works?

What we want to achieve using `awk` is to obtain the type and the value of the original keybinding of `\e[0n`. There are three types of keybindings:

1) A readline bindable function, which can be obtained by `bind -p`. This can be set up by e.g. `bind '"\e[0n": redraw-current-line'`.
2) A string macro, which can be obtained by `bind -s`. This can be set up by e.g. `bind '"\e[0n": "echo hello"'` (insert `echo hello`) or `bind '"\e[0n": "\e[C\e[C\e[C"'` (press <kbd>right</kbd> three times).
3) A unix command, which can be obtained by `bind -X`. This can be set up by e.g. `bind -x '"\e[0n": echo hello'` (print `hello`).

For the first two types, we can restore the keybinding by `bind '<spec of keybinding>'`. However, we need to specify an option `-x` for the third type to restore the keybinding: `bind -x '<spec of keybinding>'`. For this reason, in addition to the specification of the keybinding, we need to record the type of the keybinding to know how to restore the keybinding.

If we find a keybinding in `bind -ps`, we store the value to `__zoxide_original_keybinding` and set the type `__zoxide_original_keybinding_type=2` so that we can later restore the keybinding by `bind "${__zoxide_original_keybinding}"`. When we cannot find it in `bind -ps`, we then check the output of `bind -X` and, if found, store the value and the type to `__zoxide_original_keybinding` and `__zoxide_original_keybinding_type`, respectively, so that we can later restore the keybinding by `bind -x ""${__zoxide_original_keybinding}`.

Maybe it is clearer to write it using pipelines of simpler commands. I think we can rewrite it in the following way, though the following code requires about 5 extra `fork`s and 3 extra `exec`s:

```bash
__zoxide_original_keybinding=$(bind -ps | grep -E '^\\(e|M-)\[0n"' | head -n 1)
if [[ -n ${__zoxide_original_keybinding} ]]; then
    __zoxide_original_keybinding_type=2
else
    __zoxide_original_keybinding=$(bind -ps | grep -E '^\\(e|M-)\[0n"' | head -n 1)
    if [[ -n ${__zoxide_original_keybinding} ]]; then
        __zoxide_original_keybinding_type=3
    else
        __zoxide_original_keybinding_type=0
    fi
fi
```

> I'm wondering if there's a way to do it with just shell builtins (i.e. no `awk`).

There is a way, but the overhead would be large, if your concern is the spawn cost of `awk`. `bind -psX` would produce hundreds of lines, and there is no way to print only the keybinding of a specified key. It is usually not a good idea to process hundreds of lines one by one using the shell's `while` (or `for`) loop.

Nevertheless, the shell implementation would be like this:

```bash
__zoxide_original_keybinding=$(
    bind -ps | while IFS= \builtin read -r line; do
        if [[ line == '"\e[0n"'* || line ==  '"\M-[0n"'* ]]; then
            \builtin printf '%s\n' "$line"
            break
        fi
    done
)
if [[ -n ${__zoxide_original_keybinding} ]]; then
    __zoxide_original_keybinding_type=2
else
    __zoxide_original_keybinding=$(
        bind -X | while IFS= \builtin read -r line; do
            if [[ line == '"\e[0n"'* || line ==  '"\M-[0n"'* ]]; then
                \builtin printf '%s\n' "$line"
                break
            fi
        done
    )
    if [[ -n ${__zoxide_original_keybinding} ]]; then
        __zoxide_original_keybinding_type=3
    else
        __zoxide_original_keybinding_type=0
    fi
fi
```

akinomyoga:
I measured the times of processing by the difference of `EPOCHREALTIME`. The unit is in microseconds.

```
-- implementation by awk --
7781
8120
8585
6860
8164
8096
8462
8406
6254
9024
8560
8537

-- implementation by grep & head --
11553
21184
12341
25457
11910
11585
11541
12179
12057
11411
11916
12121

-- implementation by while read loop --
19017
29545
18954
19186
18920
19214
18892
19189
19286
19824
19085
45802
```

ajeetdsouza:
My main concern is more around portability than performance in this case - I was hoping to keep shell plugins free of any dependency other than zoxide. The time difference seems low enough that it would not be noticeable - thanks for profiling it!

Do you think you could write it as a pure shell implementation? Other than that, I think this should be good to merge.

---

P.S. in the example you sent, I think you may need to change `line == '"\e[0n"'*` to `"$line" == '"\e[0n"'*`.

tessus:
I hope I am not speaking out of term here, but I do have few comments after having used a command line and Unix systems since the late 80s.

> My main concern is more around portability than performance in this case - I was hoping to keep shell plugins free of any dependency other than zoxide.

This is a great goal, but in this case it should be more like a guideline or used for external tools that haven't been part of a default unix landscape for decades.

I haven't seen a unix flavored OS that did not come with awk or sed by default. Even if it were, it'd take about 2 seconds to install them. I've been writing shell scripts for decades and I have noticed that while binaries like `cut` and `sort` are almost always available on systems by default, they unfotunately use different implementations, thus the arguments are different and/or they have different features.

However, when it comes to `awk` and `sed` the story is a bit different. Also, the awk code that akinomyoga wrote is portable.

To be perfectly honest, one who does not use sed or awk on their command line most likely doesn't need zoxide either. Explanation: People who use the command line (instead of only clicking on icons and not knowing how to open a terminal) usually like to write one liners and those often involve sed or awk.

> The time difference seems low enough that it would not be noticeable

This is true if you only look at zoxide. If it all were an isolated situation. These days tools like zoxide require shell integration and initialization.

People who use zoxide add `eval $(zoxide init shell)`, if they use starship they add `eval $(starship init shell)`, and so on. You get my drift. It adds up. 
e.g. My terminal takes about 1.2s for the prompt to finally show up. I traced the perf and all the tools are very fast in isolation, but, if you use multiple tools, it just adds up.

The same is true for hooks. Many of the tools use hooks, which might trigger their integrations at the same action. So in this case we have an average of 300% worse performance when using shell only code. Ok, so the difference beteen 7000 and 23000 microsecs doesn't seem to be that bad - in isolation. But multiply that by X tools and it becomes noticeable.

Personanlly I experience a lag when an action takes more than 100 ms.

Just my 2 cents.

ajeetdsouza:
> I haven't seen a unix flavored OS that did not come with awk or sed by default.

The problem is that the installed `awk` itself may not be the one you want. This PR already includes a special-case for Solaris, because it has a different awk by default. Then there's BusyBox awk on some systems, awk on BSD systems, gawk on GNU systems, etc. Some users may have installed `goawk` for fun and aliased it. Then there's systems like Bash on Windows (Cygwin, MSYS2, Git Bash) - `awk` there is a shell script that calls `gawk`, and some users reported that it doesn't start properly without some changes. On macOS, there is the builtin awk, that gets overridden if you install awk via Homebrew.

In this case, I'd much rather minimize the number of variables than debug issues due to differing `awk` implementations.

> People who use zoxide add `eval $(zoxide init shell)`, if they use starship they add `eval $(starship init shell)`, and so on. You get my drift. It adds up. e.g. My terminal takes about 1.2s for the prompt to finally show up. I traced the perf and all the tools are very fast in isolation, but, if you use multiple tools, it just adds up.

True, but the init / hook code of zoxide is as fast as it could possibly be. The code in this PR is run only when using Space+Tab completions, so it will not affect anything else.

akinomyoga:
I'm going to adjust this PR later to use the shell implementation.

> This PR already includes a special-case for Solaris, because it has a different awk by default.

However, I have to say that all the awk implementations should be able to process the AWK script in this PR. I'm testing my AWK scripts with various awk implementations in various systems (including *BSD, Minix, Cygwin, MSYS1&2, Haiku, Solaris, etc.), but all awk implementations except for Solaris awk basically follow the POSIX standard. Solaris awk is the only exception that behaves very differently from the other awk implementations currently available. It is intentional that Solaris awk at `/usr/bin/awk` is so different; it's for backward compatibility with an ancient implementation of 1977 awk. Solaris officially provides the standard conforming version at `/usr/xpg4/bin/awk`, so in Solaris, anyone writing portable AWK scripts is supposed to use `/usr/xpg4/bin/awk` instead of `/usr/bin/awk`. Actually, I'm not sure whether Solaris `/usr/bin/awk` indeed fails to process the AWK script in this PR. I just blindly use `/usr/xpg4/bin/awk` for processing portable AWK scripts in Solaris because it seems to be the convention. So you can't use the fact that this PR has a switch for Solaris awk to justify that the AWK script in this PR may have a compatibility issue with awk implementations.

akinomyoga:
I added shell implementations and compared different implementations. The following table shows the average time in milliseconds (with statistical standard errors). The time is measured for two cases: 1) When the keybinding to `"\e[0n"` exists (With binding), and 2) when the keybinding to `"\e[0n"` does not exist (Without binding).

| impl | With binding  | Without binding |
|------|:-------------:|:---------------:|
| 1    | 8.17 ± 0.14  | 8.74 ± 0.07    |
| 2    | 8.63 ± 0.10  | 10.88 ± 0.37   |
| 3    | 13.75 ± 0.07 | 25.83 ± 2.00   |
| 4    | 13.56 ± 0.09 | 22.48 ± 0.11   |
| 5    | 10.16 ± 0.49 | 11.11 ± 0.93   |
| 6    | 6.07 ± 0.06  | 8.94 ± 0.70    |

Impl 1 is an updated version of the awk implementation (I added `LC_ALL=C`). Impl 2 is a fixed version of the grep & head implementation (I fixed the second `bind -ps` to `bind -X`). Impl 3 is a fixed version of the shell implementation (I fixed `line == ...` to `${line} == ...`, thanks to @ajeetdsouza). Impl 4 is a performance-improved variant of impl 3. Impl 5 uses Bash's `compgen`. Impl 6 is a variant of impl 5 but uses Bash 5.3's function substitutions and `compgen -V` to improve the performance. The detailed implementations can be checked in the following section.

<details><summary>Implementations</summary>

## Impl 1: awk

```bash
\builtin local awk=awk
if [[ ${OSTYPE} == solaris* && -x /usr/xpg4/bin/awk ]]; then
    # Note: The default Solaris awk is not POSIX compliant.  Solaris
    # provides the standard-conforming version at /usr/xpg4/bin/awk.
    awk=/usr/xpg4/bin/awk
fi
__zoxide_original_keybinding=$(
    {
        bind -ps || true
        \builtin printf '__unix_command__\n'
        bind -X 2>/dev/null || true
    } | LC_ALL=C "${awk}" '
        BEGIN { exit_status = 2; }
        /^__unix_command__$/ { exit_status = 3; next; }
        /^"\\(e|M-)\[0n"/ { print; exit exit_status; }
    '
)
__zoxide_original_keybinding_type=$?
```

### Impl 2: grep & head

```bash
__zoxide_original_keybinding=$(bind -ps | grep -E '^"\\(e|M-)\[0n"' | head -n 1)
if [[ -n ${__zoxide_original_keybinding} ]]; then
    __zoxide_original_keybinding_type=2
else
    __zoxide_original_keybinding=$(bind -X 2>/dev/null | grep -E '^"\\(e|M-)\[0n"' | head -n 1)
    if [[ -n ${__zoxide_original_keybinding} ]]; then
        __zoxide_original_keybinding_type=3
    else
        __zoxide_original_keybinding_type=0
    fi
fi
```

### Impl 3: while read

```bash
__zoxide_original_keybinding=$(
    bind -ps | while IFS= \builtin read -r line; do
        if [[ ${line} == '"\e[0n"'* || ${line} ==  '"\M-[0n"'* ]]; then
            \builtin printf '%s\n' "$line"
            break
        fi
    done
)
if [[ -n ${__zoxide_original_keybinding} ]]; then
    __zoxide_original_keybinding_type=2
else
    __zoxide_original_keybinding=$(
        bind -X 2>/dev/null | while IFS= \builtin read -r line; do
            if [[ ${line} == '"\e[0n"'* || ${line} ==  '"\M-[0n"'* ]]; then
                \builtin printf '%s\n' "$line"
                break
            fi
        done
    )
    if [[ -n ${__zoxide_original_keybinding} ]]; then
        __zoxide_original_keybinding_type=3
    else
        __zoxide_original_keybinding_type=0
    fi
fi
```

### Impl 4: while read (combined)

```bash
__zoxide_original_keybinding=$(
    {
        bind -ps || true
        \builtin printf '__unix_command__\n'
        bind -X 2>/dev/null || true
    } | {
        exit_status=2
        while IFS= \builtin read -r line; do
            case $line in
            (__unix_command__)
                exit_status=3 ;;
            ('"\e[0n"'* | '"\M-[0n"'*)
                \builtin printf '%s\n' "$line"
                exit "${exit_status}" ;;
            esac
        done
    }
)
__zoxide_original_keybinding_type=$?
```

### Impl 5: compgen

```bash
local REPLY IFS=$'\n'
if REPLY=$(shopt -s extglob; compgen -W '$(bind -ps)' -X '!"\\@(e|M-)\[0n"*'); then
    __zoxide_original_keybinding=${REPLY%%"$IFS"*}
    __zoxide_original_keybinding_type=2
elif REPLY=$(shopt -s extglob; compgen -W '$(bind -X 2>/dev/null)' -X '!"\\@(e|M-)\[0n"*'); then
    __zoxide_original_keybinding=${REPLY%%"$IFS"*}
    __zoxide_original_keybinding_type=3
else
    __zoxide_original_keybinding=
    __zoxide_original_keybinding_type=0
fi
```

### Impl 5: compgen (Bash 5.3)

```bash
local REPLY IFS=$'\n'
local extglob=1
shopt -q extglob || extglob=0
shopt -s extglob
if compgen -V REPLY -W '${ bind -ps; }' -X '!"\\@(e|M-)\[0n"*'; then
    __zoxide_original_keybinding=${REPLY[0]}
    __zoxide_original_keybinding_type=2
elif compgen -V REPLY -W '${ bind -X; }' -X '!"\\@(e|M-)\[0n"*'; then
    __zoxide_original_keybinding=${REPLY[0]}
    __zoxide_original_keybinding_type=3
else
    __zoxide_original_keybinding=
    __zoxide_original_keybinding_type=0
fi
if ((extglob == 0)); then
    shopt -u extglob
fi
```
</details>

I pushed a shell implementation with Impl 5 & 6 (Impl 6 for Bash >= 5.3, or impl 5 otherwise).

akinomyoga:
rebased.

ajeetdsouza:
Nice, you actually beat the performance of awk! I had an idea last night and sent a PR to fzf - figured that if we just make it a convention to set the keybinding before using it, it should work out better for everybody (especially since `\e[0n` is not really used for any other purpose). Looks like it got merged: https://github.com/junegunn/fzf/pull/4377

I think saving / restoring the keybinding should no longer be necessary now. Can we remove that and just keep the changes to bind + the fallback? I'm really sorry about the back-and-forth on this PR!

tessus:
> I think saving / restoring the keybinding should no longer be necessary now.

Well, what about when people use older versions of fzf or actually the current one (the PR was merged, but no new release has been cut)?
Also, isn't the keybinding only saved, if one was set in the first place? So by remving that code, we only make it cover less edge cases.
What about other tools? Can we be 100% sure that the keybinding was actually set (or not set) by fzf?

akinomyoga:
> Nice, you actually beat the performance of awk! I had an idea last night and sent a PR to fzf - figured that if we just make it a convention to set the keybinding before using it, it should work out better for everybody (especially since `\e[0n` is not really used for any other purpose). Looks like it got merged: [junegunn/fzf#4377](https://github.com/junegunn/fzf/pull/4377)

That breaks ble.sh's integration with Fzf...

ajeetdsouza:
Wait, why? fzf isn't using `builtin bind`.

akinomyoga:
> That breaks ble.sh's integration with Fzf...

OK, I think this can be worked around at ble.sh's side.

> What about other tools? Can we be 100% sure that the keybinding was actually set (or not set) by fzf?

I think I've seen a few times that other tools also use this technique, but I don't remember exactly. Maybe I also saw it being explained on StackOverflow, so I anticipate random dotfiles written by users may also use this. At least, I know that `skim` (whose completion framework is taken from Fzf) uses the same trick.

https://github.com/skim-rs/skim/blob/991fbee680fa464e9d9394aa925d3a087b38a426/shell/key-bindings.bash#L118

akinomyoga:
> Wait, why? fzf isn't using `builtin bind`.

OK, thinking about it again, this is probably harmless even though it would leave an unused keybinding to `\e[0n`. In ble.sh's integration with Fzf, both `bind '"\e[0n": redraw-current-line'` and `printf '\e[5n'` are suppressed by an ugly hack of temporarily replacing the builtins. With the new Fzf setting, the suppression of the former is *broken*, so the keybinding to `\e[0n` (which would never be used in the presence of ble.sh) would be set up. However, it wouldn't cause an actual problem (except that the keybinding to `\e[0n` that other tools might set up could be lost).

akinomyoga:
I reordered the commits to make the commit for saving/restoring to the last one, and then I dropped it (commit 17d521e).

ajeetdsouza:
> I have to say that all the awk implementations should be able to process the AWK script in this PR.

I agree, it would be extremely unlikely for an awk implementation to not work with the code in your earlier PR. However, due to the number of users of zoxide, I've spent a lot of time debugging very strange issues only to find out that it was caused by something in the user's setup. I know it looks like an odd request, but the number of these issues I've encountered is why all the code in the shell plugin is so defensively written, and it's also why I was in favor of a pure shell implementation: https://asq.org/quality-resources/mistake-proofing

> Well, what about when people use older versions of fzf or actually the current one (the PR was merged, but no new release has been cut)?

I suppose the same question could be asked about users on older versions of zoxide - this is a bugfix, you'll need to update your software to get it. I think as a convention, it's a good idea to `bindkey \e[0n` before using it - especially since this is the only known way of calling readline functions and it is non-trivial to restore the keybinding once you're done.

ajeetdsouza:
Merged, thank you so much @akinomyoga, and sorry about the confusion!

I am now a happy user of ble.sh :)

akinomyoga:
Thanks!

ajeetdsouza:
> I thought that this fix actually switches into the dir automatically.

Nope, to the best of my knowledge there is no way to do that. I tried a couple of things:

- I tried to directly `cd` into the directory from the completion function, but there doesn't seem to be a good way to redraw the prompt once the fzf window closes. So while the directory gets changed, the prompt (which usually shows the current directory) doesn't get updated to the new directory.
- After replacing the current readline buffer with `cd /path/here`, I also tried looking for a way to submit the command to Bash, but there again I couldn't find anything.

Maybe @akinomyoga has some ideas, but I don't know of any way to make the autocompleter run the `cd` command for you in the same way as zsh or fish.

akinomyoga:
> * So while the directory gets changed, the prompt (which usually shows the current directory) doesn't get updated to the new directory.

Yes, this is the problem. Bash evaluates `PS1` only once for each command. Until a new command is executed from the command line, `PS1` is not evaluated again. The previous evaluation result (which contains the previous directory) is continued to be shown until the next command execution.

> * After replacing the current readline buffer with `cd /path/here`, I also tried looking for a way to submit the command to Bash, but there again I couldn't find anything.

I haven't tried, but I think one can try `bind '"\e[0n": accept-line'; printf '\e[5n'` in `__zoxide_z_complete_helper`.

ajeetdsouza:
> I haven't tried, but I think one can try `bind '"\e[0n": accept-line'; printf '\e[5n'` in `__zoxide_z_complete_helper`.

I did try this, but Bash will run the `accept-line` function before the line actually gets changed, so it just ends up running the command the user types in as-is.

tessus:
Thanks guys. Too bad, but I knew that bash has its limits. I just hoped there was some sort of a workaround. ;-)

Anyway, thanks again for all the work that went into this!

akinomyoga:
@ajeetdsouza As I mentioned in https://github.com/ajeetdsouza/zoxide/pull/1048#issuecomment-2868567922, Skim has the integration code derived from that of Fzf, which [binds `"\e[0n": redraw-current` globally](https://github.com/skim-rs/skim/blob/991fbee680fa464e9d9394aa925d3a087b38a426/shell/key-bindings.bash#L118). This needs to be fixed the same way as you modified fzf at https://github.com/junegunn/fzf/pull/4377. Could you apply and submit the same patch to Skim?

ajeetdsouza:
Done: https://github.com/skim-rs/skim/pull/774

akinomyoga:
Thank you!

akinomyoga:
> https://github.com/ajeetdsouza/zoxide/pull/1048#issuecomment-2868567922
>
> > That breaks ble.sh's integration with Fzf...
> 
> OK, I think this can be worked around at ble.sh's side.

FYI, I added the workaround in https://github.com/akinomyoga/blesh-contrib/compare/45b477ed509bb254c9efb0882740a741cb83e107...b4696164f098c557f51e97cf8036e703524a256b.

PR #1052, opened 2025-05-13
Add PowerShell ArgumentCompleter

### LINKED ISSUE #344: Tab completions for PowerShell

twobiers:
Resolves #344 ,
Reopens #419

I have tried to implement an equivalent Tab completion for PowerShell, which was to some point successful. 
It works like the following:
- Use the default directory completion when no arguments are provided
- If an argument is provided, use it as zoxide query input

Unfortunately it doesn't seem to be possible to run interactive queries in PowerShell within the argument completer, therefore I don't know whether it is possible to implement the "space+tab" completion. 
Tested on Windows 11 in PowerShell just today.

This GIF shows, how it works:
WindowsTerminal_caWG2ym896

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -144,6 +144,52 @@ function global:__zoxide_zi {
 Set-Alias -Name {{cmd}} -Value __zoxide_z -Option AllScope -Scope Global -Force
 Set-Alias -Name {{cmd}}i -Value __zoxide_zi -Option AllScope -Scope Global -Force
 
+filter __zoxide_escapeStringWithSpecialChars {
+    $_ -replace '\s|#|@|\$|;|,|''|\{|\}|\(|\)|"|`|\||<|>|&','`$&'
+}
+
+Register-ArgumentCompleter -Native -CommandName "{{cmd}}" -ScriptBlock {
+    param(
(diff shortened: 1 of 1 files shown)


benzaria:
> Thanks! I think it may be possible to add interactive completions as well, but I'll need a week or two till I'm able to get my hands on a Windows machine to test this.

u may use pwsh on any system, no 🤔

ajeetdsouza:
> u may use pwsh on any system, no 🤔

That is correct, but the terminal on Windows doesn't support a lot of the features that most macOS / Linux terminals do, so it's usually advisable to test out features like this on both systems.

ajeetdsouza:
Not sure why, but this is somehow not working on Windows Terminal / PowerShell 7.5.1 - I'm just getting normal directory completions. Any idea why this could be happening?

ajeetdsouza:
Closing due to lack of a response.

PR #1053, opened 2025-05-13
Some improvements for powershell

benzaria:
- Optimizing: `__zoxide_z` function for better preformance.
- Modifying: `cd +` to look for latest made directory and switch to It, in case of no found next Dirs.
- Adding: `cd ++` a Forced vesion of `cd +` that only look for latest created Dir and Switch Location.

The new `cd +` can very useful when working with filesystem for fast directory making and switching and makes zoxide more smart🤓


ajeetdsouza:
Hey, thanks for the PR. A couple of pointers:

> * Optimizing: `__zoxide_z` function for better preformance.

The only thing that seems to have changed here is the way the conditions are computed. I don't think this will have any impact on the performance of the `__zoxide_z` function. Unless you have profiled the function and found a measurable improvement in performance, this seems like an unnecessary change.

> * Modifying: `cd +` to look for latest made directory and switch to It, in case of no found next Dirs.
> * Adding: `cd ++` a Forced vesion of `cd +` that only look for latest created Dir and Switch Location.

While this is an interesting feature, it also seems very niche - most people don't really want their cd command to have this much of a learning curve. I try to keep features restrictive so that zoxide feels intuitive, and I don't think this feature comes under that umbrella. Additionally, this PR would make the PowerShell plugin inconsistent with all the other shells supported by zoxide.

---

I'm closing this for now, but as a rule of thumb, when adding features to a repository, please create an issue first! The maintainers might have a different approach that you may want to consider, or they may not find the feature a good fit for the project.

PR #1062, opened 2025-05-23
Update powershell.txt to solve uncatched hook action error in the powershell hook

Zim-Inn:
Thie change can avoid zoxide hook to influence user's handling of $Error varible. 
For example, if it is still unhandled, the zoxide init error will be propagated to user's  scope. 
Thus it will make user's script can't work as their expects.

### DIFF
--- a/templates/powershell.txt
+++ b/templates/powershell.txt
@@ -80,7 +80,12 @@ function global:__zoxide_hook {
 {%- endif %}
 
 # Initialize hook.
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Wasn't this solved by https://github.com/ajeetdsouza/zoxide/pull/1038?

Zim-Inn:
> Wasn't this solved by #1038?

Yes, i confirm that it also did resolve it .
You can close the PR.

PR #1064, opened 2025-05-24
Provide completions for Nushell

ysthakur:
This PR provides completions for `z` and `zi`. Nushell no longer uses external completers for internal commands since v0.103 ([release notes](https://www.nushell.sh/blog/2025-03-18-nushell_0_103_0.html#external-completers-are-no-longer-used-for-internal-commands-toc)), so people's old external Zoxide completers won't work anymore.

This completer uses the option `completion_algorithm: "substring"`, which was only added in 0.104. An alternative would be to use `completion_algorithm: "prefix", positional: false`, which works in older versions but is [deprecated](https://www.nushell.sh/blog/2025-04-29-nushell_0_104_0.html#substring-match-algorithm-toc).

This completer also looks a little weird if you type in multiple arguments. If you type in `z /foo bar<TAB>`, all the completions you see will have everything up to and including the first `/foo` stripped. It's not great, but it's needed because custom completers in Nushell can't currently choose which span they want to replace. Without it, if your first result upon typing `z /foo bar<TAB>` were `/foo/bar`, hitting Enter would replace your command line with `z /foo /foo/bar`, which is incorrect.

[asciicast](https://asciinema.org/a/cKElGuomdJVl7IF1ywrOLO7cE?t=7)

### DIFF
--- a/templates/nushell.txt
+++ b/templates/nushell.txt
@@ -55,29 +55,63 @@ export-env {
 # When using zoxide with --no-cmd, alias these internal functions as desired.
 #
 
-# Jump to a directory using only keywords.
-def --env --wrapped __zoxide_z [...rest: string] {
-  let path = match $rest {
-    [] => {'~'},
-    [ '-' ] => {'-'},
-    [ $arg ] if ($arg | path expand | path type) == 'dir' => {$arg}
-    _ => {
-      zoxide query --exclude $env.PWD -- ...$rest | str trim -r -c "\n"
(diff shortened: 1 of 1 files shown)


ysthakur:
@ajeetdsouza Could you take a look at this?

azaleacolburn:
@ajeetdsouza This would be greatly appreciated.

ysthakur:
Superseded by https://github.com/ajeetdsouza/zoxide/pull/1129

PR #1072, opened 2025-06-05
Provide jump commands completion for nushell

alecandido:
For custom commands (defined in Nushell itself), completion has to be directly attached to the command definition
https://www.nushell.sh/book/custom_completions.html#custom-completions

So, better to provide completion in the template itself :)

---

Notice that there is an entire section in the official docs of Nushell discussing zoxide completion
https://www.nushell.sh/cookbook/external_completers.html#zoxide-completer

However,
> This completer is not usable for almost every other command, so it's recommended to add it as an override in the [multiple completer](https://www.nushell.sh/cookbook/external_completers.html#multiple-completer):
> [...]
> Sometimes, a single ***external*** completer is not flexible enough.

i.e., it always assumes that zoxide commands are provided as external ones. While the Nushell template is defining `__zoxide_z` and `__zoxide_zi` as custom commands (internal ones).

### DIFF
--- a/templates/nushell.txt
+++ b/templates/nushell.txt
@@ -55,8 +55,25 @@ export-env {
 # When using zoxide with --no-cmd, alias these internal functions as desired.
 #
 
+# Provide suitable completion for zoxide jump commands
+def zoxide_completer [context: string] {
+      let query = $context | split row " " | slice 1..
+      let matches = zoxide query -l ...$query | lines | where {|x| $x != $env.PWD}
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Hey, thanks for sending this in. For consistency and ergonomics, shell completions should be based off of `zoxide query -i` - closing in favor of https://github.com/ajeetdsouza/zoxide/pull/1129.

PR #1078, opened 2025-06-16
Update README.md for RHEL 10 & derivatives

nisbet-hubbard:
The Fedora Rust SIG has indicated that Zoxide won’t be branched for EPEL 10: https://bugzilla.redhat.com/show_bug.cgi?id=2371051#c2.

The affected platforms are RHEL 10+ and its derivatives. I’ve updated the README to reflect this.


ajeetdsouza:
Thanks!

PR #1080, opened 2025-06-17
Add support for multi digits dirstack entry (e.g. +12).

### LINKED ISSUE #1074: DIRSTACK usage "z +n" and "z -n" fails with double digits in n
When dirstack has more than 10 entries, the command `z +1` ... `z +9` works, but `z +11` does not work.

> zoxide: no match found

Tony-Busy:
In zsh, a dirstack entry with multiple digits will fail even if the dirstack entries are more than that index.
```
❯ dirs -v
0    ~/
1    ~/XXX/YY
...
11    ~/ZZ/CC
12    ~/AAA
...
❯ z +12
zoxide: no match found
```
This corrects this failure by allowing a multi-digit entry index.

Fixes #1074

### DIFF
--- a/templates/zsh.txt
+++ b/templates/zsh.txt
@@ -84,7 +84,7 @@ function __zoxide_z() {
     __zoxide_doctor
     if [[ "$#" -eq 0 ]]; then
         __zoxide_cd ~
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Thanks!

PR #1100, opened 2025-08-11
feat: allow relative paths for --base-dir

tom-anders:
Use std::filesystem::canonicalize to transform to an absolute path.


azaleacolburn:
This was an oversight with my original PR, I support this change.

PR #1105, opened 2025-08-18
Unpin and update tempfile

musicinmybrain:
No longer pin an exact version of the `tempfile` dev-dependency.

The pin was added in https://github.com/ajeetdsouza/zoxide/commit/b318a2a190096ffdbeb5a508e95f57332b0f2a13 without comment or rationale; there is no obvious reason to think that later SemVer-compatible releases should not work.

After removing the exact-version pin, run `cargo update tempfile` to update `Cargo.lock` with the latest release.

I confirmed that `cargo test` still passes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: install
-          args: --color=always --git=https://github.com/cross-rs/cross.git --locked --rev=02bf930e0cb0c6f1beffece0788f3932ecb2c7eb --verbose cross
+          args: --color=always --git=https://github.com/cross-rs/cross.git --locked --rev=e281947ca900da425e4ecea7483cfde646c8a1ea --verbose cross
       - name: Build binary
         uses: actions-rs/cargo@v1
         with:


ajeetdsouza:
See the CI failure - that's the main reason `tempfile` was pinned. There is some kind of Cross failure with the latest version of getrandom that I haven't been able to look into yet.

musicinmybrain:
> See the CI failure - that's the main reason `tempfile` was pinned. There is some kind of Cross failure with the latest version of getrandom that I haven't been able to look into yet.

Looking at https://github.com/cross-rs/cross/issues/1619, it appears that this should be fixed in Cross with https://github.com/cross-rs/cross/commit/107433fa9d71d7eb506cd5bf5467bab89edb7a52. This project currently has https://github.com/cross-rs/cross/commit/02bf930e0cb0c6f1beffece0788f3932ecb2c7eb, which is older:

https://github.com/ajeetdsouza/zoxide/blob/c070535968e04b1b34ff4d025162f5f68fdd7d69/.github/workflows/release.yml#L70

I suppose there is no reason not to try the latest commit https://github.com/cross-rs/cross/commit/e281947ca900da425e4ecea7483cfde646c8a1ea; let me know if you know of a reason to try a different commit instead. I’m going to modify the hash in `.github/workflows/release.yml` and see what happens.

musicinmybrain:
It looks like that worked as I hoped it would!

ajeetdsouza:
This is great, thank you! 😃

PR #1106, opened 2025-08-23
Update Nushell completion

### LINKED ISSUE #1098: Nushell integration broken since 0.103
Just reporting that the generated script no longer works in nushell 0.103.

As per this blog post : https://www.nushell.sh/blog/2025-03-18-nushell_0_103_0.html#external-completers-are-no-longer-used-for-internal-commands-toc

I found this workaround script which works well for me : https://github.com/Jan9103/nu-snippets/blob/main/integration/zoxide.nu

It basically needs to be a function definition with a linked completer instead of a generic external completer. 

This makes it work well with other generic external completers such as carapace without having to build a custom external completer that switches between which completer is wanted based on the command.

beicause:
Fixes #1098.

Nushell [external completers are no longer used for internal commands](https://www.nushell.sh/blog/2025-03-18-nushell_0_103_0.html#external-completers-are-no-longer-used-for-internal-commands-toc) so we should include the completion for commands.

### DIFF
--- a/templates/nushell.txt
+++ b/templates/nushell.txt
@@ -51,12 +51,25 @@ export-env {
 
 {%- endif %}
 
+def "nu-complete zoxide path" [context: string] {
+  let parts = $context | split row " " | skip 1
+  {
+    options: {
+      sort: false
+      completion_algorithm: prefix
+      positional: false
+      case_sensitive: false
+    }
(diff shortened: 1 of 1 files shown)


beicause:
There are already PRs to do this, see #1072 and #1064.

PR #1108, opened 2025-08-28
Strikeout apt.cli.rs due to sunsetting

halilim:
It's sunsetting: https://github.com/emmatyping/apt.cli.rs/issues/32


ajeetdsouza:
Thanks!

PR #1111, opened 2025-09-07
Adds Murex shell support

JeffMelton:
Murex is niche, but I use it, and I miss zoxide. 😄

---
<a href="https://gitkraken.dev/review/github/ajeetdsouza/zoxide/pull/1111?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/contrib/completions/_zoxide
+++ b/contrib/completions/_zoxide
@@ -178,7 +178,7 @@ _arguments "${_arguments_options[@]}" : \
 '--help[Print help]' \
 '-V[Print version]' \
 '--version[Print version]' \
-':shell:(bash elvish fish nushell posix powershell tcsh xonsh zsh)' \
+':shell:(bash elvish fish murex nushell posix powershell tcsh xonsh zsh)' \
 && ret=0
 ;;
 (query)
--- a/contrib/completions/zoxide.bash
+++ b/contrib/completions/zoxide.bash
@@ -275,7 +275,7 @@ _zoxide() {
             return 0
             ;;
         zoxide__subcmd__init)
-            opts="-h -V --no-cmd --cmd --hook --help --version bash elvish fish nushell posix powershell tcsh xonsh zsh"
+            opts="-h -V --no-cmd --cmd --hook --help --version bash elvish fish murex nushell posix powershell tcsh xonsh zsh"
             if [[ ${cur} == -* || ${COMP_CWORD} -eq 2 ]] ; then
                 COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                 return 0
--- a/contrib/completions/zoxide.nu
+++ b/contrib/completions/zoxide.nu
@@ -93,7 +93,7 @@ module completions {
   ]
 
   def "nu-complete zoxide init shell" [] {
-    [ "bash" "elvish" "fish" "nushell" "posix" "powershell" "tcsh" "xonsh" "zsh" ]
+    [ "bash" "elvish" "fish" "murex" "nushell" "posix" "powershell" "tcsh" "xonsh" "zsh" ]
   }
 
   def "nu-complete zoxide init hook" [] {
--- a/contrib/completions/zoxide.ts
+++ b/contrib/completions/zoxide.ts
@@ -285,6 +285,7 @@ const completion: Fig.Spec = {
           "bash",
           "elvish",
           "fish",
+          "murex",
           "nushell",
           "posix",
           "powershell",
--- a/shell.nix
+++ b/shell.nix
@@ -25,6 +25,17 @@ in pkgs.mkShell {
     pkgs.elvish
     pkgs.fish
     pkgs.ksh
+    (pkgs.murex.overrideAttrs (old: rec {
+      version = "7.2.1001";
+      src = pkgs.fetchFromGitHub {
+        owner = "lmorg";
+        repo = "murex";
+        rev = "v${version}";
+        hash = "sha256-Ua5KEtT1HXRCqW4MwB0dYCd03DBrliEfgiSmcp+vZS8=";
+      };
+      vendorHash = "sha256-MaBBi2Qi7s9lfRWmnYkyr7PtwzC7ZL0jmyUXzISOXVg=";
+      doCheck = false;
+    }))
     pkgs.nushell
     pkgs.powershell
     pkgs.tcsh
--- a/src/cmd/cmd.rs
+++ b/src/cmd/cmd.rs
@@ -156,6 +156,7 @@ pub enum InitShell {
     Bash,
     Elvish,
     Fish,
+    Murex,
     Nushell,
     #[clap(alias = "ksh")]
     Posix,
--- a/src/cmd/init.rs
+++ b/src/cmd/init.rs
@@ -6,7 +6,7 @@ use askama::Template;
 use crate::cmd::{Init, InitShell, Run};
 use crate::config;
 use crate::error::BrokenPipeHandler;
-use crate::shell::{Bash, Elvish, Fish, Nushell, Opts, Posix, Powershell, Tcsh, Xonsh, Zsh};
+use crate::shell::{Bash, Elvish, Fish, Murex, Nushell, Opts, Posix, Powershell, Tcsh, Xonsh, Zsh};
 
 impl Run for Init {
     fn run(&self) -> Result<()> {
@@ -19,6 +19,7 @@ impl Run for Init {
             InitShell::Bash => Bash(opts).render(),
             InitShell::Elvish => Elvish(opts).render(),
             InitShell::Fish => Fish(opts).render(),
+            InitShell::Murex => Murex(opts).render(),
             InitShell::Nushell => Nushell(opts).render(),
             InitShell::Posix => Posix(opts).render(),
             InitShell::Powershell => Powershell(opts).render(),
--- a/src/shell.rs
+++ b/src/shell.rs
@@ -26,6 +26,7 @@ macro_rules! make_template {
 make_template!(Bash, "bash.txt");
 make_template!(Elvish, "elvish.txt");
 make_template!(Fish, "fish.txt");
+make_template!(Murex, "murex.txt");
 make_template!(Nushell, "nushell.txt");
 make_template!(Posix, "posix.txt");
 make_template!(Powershell, "powershell.txt");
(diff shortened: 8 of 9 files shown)


ajeetdsouza:
Hi, I've added some minor bugfixes and made it more consistent with the other shells. There's a few more things I'd like to fix before merging:

- PWD hooks are possible to implement here. What you need to do is use a global variable called `__zoxide_oldpwd`, and set up a prompt hook that compares `__zoxide_oldpwd` with `$PWD`. If it has changed, set `__zoxide_oldpwd=$PWD` and call `zoxide add $PWD` (since the directory has changed).
- `z` (with no arguments) does not work.
- If you zoxide has no result for your query (e.g. `z aoiejfaoiejf`), you get a full error trace with a line number. Ideally, we should just `return 1` in such a situation.

Do you think you could work on these?

JeffMelton:
Sure can! Thanks for reviewing. 😁

On Sep 7, 2025, at 12:55 AM, ajeetdsouza ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/1777663?s=20&v=4]ajeetdsouza left a comment (ajeetdsouza/zoxide#1111)<https://github.com/ajeetdsouza/zoxide/pull/1111#issuecomment-3263510932>

Hi, I've added some minor bugfixes and made it more consistent with the other shells. There's a few more things I'd like to fix before merging:

  *   PWD hooks are possible to implement here. What you need to do is use a global variable called __zoxide_oldpwd, and set up a prompt hook that compares __zoxide_oldpwd with $PWD. If it has changed, set __zoxide_oldpwd=$PWD and call zoxide add $PWD (since the directory has changed).
  *   z (with no arguments) does not work.
  *   If you zoxide has no result for your query (e.g. z aoiejfaoiejf), you get a full error trace with a line number. Ideally, we should just return 1 in such a situation.

Do you think you could work on these?

JeffMelton:
All right, @ajeetdsouza, thanks for the feedback. I think I've covered your requests. Tests pass here, but I see the ubuntu CI test is still failing. I don't have `nix-shell` set up, but I'm trying to replicate as much as I can in a container environment. Do please let me know if you'd like to see anything else.

JeffMelton:
~Ah, I've found a flaw in my implementation. I'll update again here once I think I've sorted it.~ Sorted (I think 😆 ). ✅

ajeetdsouza:
Tried `zoxide init murex -> source` with murex v7.0.2107, but the hook does not seem to be working. A way to see if the hook is working is to do something like this:

```shell
zoxide query --score /tmp
z /tmp
# the score should have increased
zoxide query --score /tmp
# when in the same directory, the score should stay the same
zoxide query --score /tmp
```

Also, `z` with no arguments still doesn't seem to work for me:

```
Error in `__zoxide_z` (<stdin> 1,2637):
      Command: __zoxide_z  @PARAMS
      Error: array '@PARAMS' is empty
```

JeffMelton:
All right, thanks. I'll keep poking at it.

JeffMelton:
It was my understanding (perhaps incorrect) that I'd need to use the zoxide `--prompt` flag so that it would update the score if the hook picked up a directory change. That's working as expected here:
```
 🏡
μ →  murex --version
murex v7.0.2129 (website)
GPL v2
2018-2025 Laurence Morgan
μ →  /bin/tail -1 ~/.murex_profile
zoxide init murex --hook prompt -> source
μ →  zoxide query --score /tmp
zoxide: no match found
Error in `zoxide` (0,1): exit status 1
 🏡
μ →  z /tmp
/tmp
μ →  zoxide query --score /tmp
   4.0 /tmp
/tmp
μ →  z down
…/Downloads
μ →  zoxide query --score /tmp
   8.0 /tmp
…/Downloads
μ →  z tmp
/tmp
μ →  zoxide query --score /tmp
  12.0 /tmp
```
Have I misunderstood how that's supposed to work?

ajeetdsouza:
The hook is basically designed to increment the score of a directory (using `zoxide add`), so that zoxide can keep track of how frequently that directory is being used.

- `--hook=prompt` means that `zoxide add $PWD` is called every time the prompt is rendered (i.e. the user has run a command in their interactive shell).
- `--hook=pwd` means that `zoxide add $PWD` is called every time the `$PWD` changes (i.e. the user has used `cd` or `z` to change their working directory). This is the default, because it:
  a. produces better results
  b. is more efficient, since it avoids running an external command (`zoxide`) on every shell prompt

Hope that clarifies things! Let me know if you have more questions, I'd be happy to help.

JeffMelton:
Okay, thanks, that does clarify. Mine is "working"/updating the score because it's overriding default, which means the PWD hook still isn't working right.

JeffMelton:
Okay, I *think* that's got it. Is that Ubuntu CI test significant? It reads like a linter issue, but I don't see anything I recognize as actionable. Let me know if you spot anything else in any case. Happy to keep iterating.

ajeetdsouza:
Hey! Sorry for the delay here, I haven't had time to look into the bug.

The issue seems to be that something in the Murex startup sequence generates a warning when used with Nix. Since Murex isn't very popular, I'm not sure if this will be fixed any time soon.

In the meantime, I've modified the test in `shell.rs` to just check the exit code of `murex -c` and ignore stdout / stderr.

The problem with this is that `zoxide init murex -> source` always exits with a return code of `0` - `source` doesn't seem to exit with an error code even if there was a problem initializing the script. If we can find an idiomatic way (murex unit test, string comparison, anything) of getting `murex -c` to exit with an error code when there is a syntax error in zoxide's script, I think this would be good to merge.

JeffMelton:
Hey @ajeetdsouza — circling back on this with what I hope is the last meaningful update before merge review. No rush on your end; just consolidating everything in one place so when you do have time it's a short read.

## TL;DR

- `murex -c <source>` exits non-zero on broken init. Empirically proven (transcript below).
- All four behaviors you flagged previously now work cleanly: PWD hook, `z` (no args), `z <no-match>`, `zi` (no args).
- One small Nix-side change: pinned Murex to `v7.2.1001` via a surgical `overrideAttrs` (only this package, nixpkgs pin untouched). Background and rationale below.
- 525/525 tests + lint green.

## Behavioral demos

All transcripts captured against `murex v7.2.1001`, rendered init sourced cleanly, no warnings.

### PWD hook (your original repro)

```
tmp » zoxide query --score /tmp
zoxide: no match found
tmp » cd /tmp
tmp » zoxide query --score /tmp
   4.0 /tmp
```

Score increments correctly on directory change via the default `--hook=pwd` path. The hook fires from a `event onPrompt name=command-completion` handler (v7.0+ feature — more on that below).

### `z` with no arguments

```
tmp » z
~ » pwd
/root
```

The wrapper now checks `$PARAMS` arity (as JSON) before expanding `@PARAMS`, since expanding an empty `@PARAMS` errors before the function body runs. This was the bug behind your `array '@PARAMS' is empty` trace.

### `z <no-match>`

```
tmp » z aoiejfaoiejf
tmp »
```

Silent return 1, no stack trace.

### `zi` with no arguments

Fzf launches against the current DB; selection cd's. Same arity-handling fix had to be applied inside `__zoxide_zi` itself — the existing `trypipe <!null>` was swallowing the `@PARAMS` error and quietly suppressing fzf launch.

## Merge gate: `murex -c` exits non-zero on broken init

This was your stated bar for merging. To prove it directly, I broke the template in two different ways and ran the existing `murex_murex` test suite:

**Test 1 — invalid event interrupt name:**
```
$ sed -i 's/command-completion/not-a-real-interrupt/' templates/murex.txt
$ cargo nextest run --features nix-dev shell::tests::murex_murex
...
Summary: 24 tests failed, all murex_murex_* variants
```

**Test 2 — outright syntax error:**
```
$ echo 'function {{{ }}}' >> templates/murex.txt
$ cargo nextest run --features nix-dev shell::tests::murex_murex
...
Summary: 24 tests failed
```

So the existing `murex -c <source>` exit-code check catches both syntax errors and runtime registration errors during sourcing. That's broader than the strict "syntax error" framing in your earlier comment — anything that fails at source-time trips it.

(Bug-honesty caveat below.)

## The Nix pin bump (and why it's safe)

Working through this, I hit a snag: the rendered init uses `event onPrompt name=command-completion`, which is a Murex **v7.0+** feature. The `nixpkgs` revision pinned by `shell.nix` ships `murex v6.4.2063`, where `command-completion` isn't a registered interrupt, so the init errored during source under CI.

You're on v7.0.2107 per your September comment, so this isn't a new user-facing constraint — it just brings the CI environment in line with what real users have.

I tried two approaches:

1. **Bump the whole nixpkgs pin** — too invasive, ripples through every other shell.
2. **Surgical override on just the `murex` derivation** — preferred. The file already uses the `pkgs.murex.overrideAttrs (old: { doCheck = false; })` idiom, so I just extended it:

```nix
(pkgs.murex.overrideAttrs (old: rec {
  version = "7.2.1001";
  src = pkgs.fetchFromGitHub {
    owner = "lmorg";
    repo = "murex";
    rev = "v${version}";
    hash = "sha256-Ua5KEtT1HXRCqW4MwB0dYCd03DBrliEfgiSmcp+vZS8=";
  };
  vendorHash = "sha256-MaBBi2Qi7s9lfRWmnYkyr7PtwzC7ZL0jmyUXzISOXVg=";
  doCheck = false;
}))
```

One package, one override, all other shells unaffected. v7.2.1001 is the current stable release.

If you'd rather not bump Murex and prefer a v6-compatible template, the alternative is to drop `event onPrompt name=command-completion` and use an older hook mechanism (likely `before`/`after` interrupts), at the cost of more complex hook logic. Happy to do that instead if you'd prefer — just let me know.

## Honest caveats

A few things worth surfacing before you spend time on this:

1. **The test's coverage is shallower than ideal.** It catches errors that fire *during source*. It does **not** catch runtime issues at the call site — for example, the empty-`@PARAMS` bug you saw originally would still slip past the test, because `__zoxide_z` doesn't get invoked just by sourcing. I caught the remaining instances of that pattern interactively. If you want stronger coverage, we could extend the test to actually invoke `z` and `zi` after sourcing, but that's incremental work and I didn't want to scope-creep this PR.

2. **The Nix startup warning you mentioned** — I poked at this. Inside `nix-shell --pure`, fzf's preview command (`command -p ls ...`, set by `src/util.rs`) fails because there's no `/bin/ls` in pure Nix. This affects **every** shell init under nix-shell (bash, zsh, fish, all of them), not just Murex. It's an artifact of the CI environment, not a Murex-specific issue. In a normal Linux/macOS environment it works fine.

3. **The Murex startup warning under Nix** that you specifically noted in your Nov comment — I wasn't able to reproduce a warning at startup under v7.2.1001 in the project's nix-shell. If you can share the exact warning text you saw, I'm happy to dig further.

## What changed since the last review

- `templates/murex.txt` — empty-`@PARAMS` handling in both the `z`/`zi` wrappers and inside `__zoxide_zi`; comments explaining the arity-branch idiom.
- `shell.nix` — surgical Murex pin to v7.2.1001 as described above.
- Commit signed under my personal identity (jeff@melton.cloud).

Single commit (`Add Murex shell support`), rebased clean on master.

Ready for re-review whenever you have bandwidth — no rush. Thanks for the patience on this one.

PR #1118, opened 2025-09-22
Update Zoxide shell setup for Nushell in README

Rikiub:
**Auto-load** files support was added on **Nushell** [v0.96.0](https://www.nushell.sh/blog/2024-07-23-nushell_0_96_0.html), allowing `nu` files to be **auto-loaded** on shell start. This behavior is documented [here](https://www.nushell.sh/book/configuration.html#configuration-overview).

This **PR** replaces the need of have a separate file and source it *"manually"*. Instead, **Zoxide** completions now can be added with one-line command.

Also, **Zoxide** now requires [v0.96.0+](https://www.nushell.sh/blog/2024-07-23-nushell_0_96_0.html) as minimum  **Nushell** version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,22 +247,14 @@ zoxide can be installed in 4 easy steps:
    <details>
    <summary>Nushell</summary>
 
-   > Add this to the <ins>**end**</ins> of your env file (find it by running `$nu.env-path`
-   > in Nushell):
+   > Add this to the <ins>**end**</ins> of your config file (find it by running `$nu.config-path` in Nushell):
    >
    > ```sh
-   > zoxide init nushell | save -f ~/.zoxide.nu
-   > ```
-   >
-   > Now, add this to the <ins>**end**</ins> of your config file (find it by running
-   > `$nu.config-path` in Nushell):
-   >
-   > ```sh
-   > source ~/.zoxide.nu
+   > zoxide init nushell | save -f ($nu.user-autoload-dirs.0)/zoxide.nu
    > ```
    >
    > **Note**
-   > zoxide only supports Nushell v0.89.0+.
+   > zoxide only supports Nushell v0.96.0+.
 
    </details>
 
--- a/README.md
+++ b/README.md
@@ -242,7 +242,8 @@ zoxide can be installed in 4 easy steps:
    <details>
    <summary>Nushell</summary>
 
-   > Add this to the <ins>**end**</ins> of your config file (find it by running `$nu.config-path` in Nushell):
+   > Add this to the <ins>**end**</ins> of your config file.
+   > Find it by running `$nu.config-path` in Nushell:
    >
    > ```sh
    > zoxide init nushell | save -f ($nu.user-autoload-dirs.0)/zoxide.nu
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Unfortunately, this wouldn't work.

The reason zoxide requires you to add `zoxide init nushell | save -f ~/.zoxide.nu` to your Nushell env file is so that the init script is regenerated every time you start a new shell. The advantage of this approach is that every time you update zoxide, the init script gets updated.

Given how often Nushell introduces breaking changes, it is important to make this process smooth. The only other option would be to expect the user to remember to re-run the init script every time they update zoxide, which would be bad UX.

Rikiub:
Thanks for the advice, I made some changes.

Now requires write the command inside `config.nu`, so each time the shell is started, the file `autoload/zoxide.nu` is created.

I think should be enough @ajeetdsouza

ajeetdsouza:
Hey! One more thing that just occurred to me: what happens to users who don't have any autoload directories?

Rikiub:
I didn't think on it, but what about tell the users to ensure `autoload` folder is created?

---

I could add something like:

```
Before all, ensure the "autoload" directory exists. You can create it running this command:

mkdir $nu.user-autoload-dirs.0
```

Then follow the next instructions.

---

Or maybe just add both to config:

```
mkdir $nu.user-autoload-dirs.0
zoxide init nushell | save -f ($nu.user-autoload-dirs.0)/zoxide.nu
```

ajeetdsouza:
Hm, in that case, I'm not sure that this is a better way to initialize zoxide on Nushell. What do you think? 

Also, to clarify - there could be two problems here:
- the autoload directory does not exist
- the user has disabled autoloading by making autoload directories an empty list

Rikiub:
I don't think casual users would replace autoload directories, by default nushell always set one autoload directory under `~/.config/nushell/autoload` so should be fine. Advanced users at this point should know what they are doing.

My reason to use autoload directories is because is cross-platform. `$nu.user-autoload-dirs.0` will always point to system config directory, so it make things more organized.

In my opinion, the second way is the most easy for new users:

```
mkdir $nu.user-autoload-dirs.0
zoxide init nushell | save -f ($nu.user-autoload-dirs.0)/zoxide.nu
```

---

Tell me if you agree so I can send another commit 👋

DanielAdari:
Just to add some context to this conversation (as a fellow nushell enthusiast 😀 ), this is how the Starship prompt instruct installation:

https://starship.rs/guide/#step-2-set-up-your-shell-to-use-starship

```sh
mkdir ($nu.data-dir | path join "vendor/autoload")

starship init nu | save -f ($nu.data-dir | path join "vendor/autoload/starship.nu") 
```

To me this seems pretty clean :)

ajeetdsouza:
The current Starship initialization method would require users to manually re-run that command on every Starship update. If they don't remember to do so, they are likely to run into weird bugs as the CLI and init script go out-of-sync.

The current zoxide setup is just 2 lines long and doesn't seem like it has any disadvantages, so I'm closing this. Thanks!

PR #1121, opened 2025-10-03
Add recursive flag for add and remove commands

### LINKED ISSUE #1087: Add these 2 features in zoxide, 1. Add multiple directories path, 2. Remove multiple paths
I deleted zoxide 2 times because it was not fast and i watched some youtube videos and found out i don't have any paths in zoxide database so i added all paths of those directories i regularly visit's using zpath_add and removed some paths that are not important using zpath_remove.

Now my zoxide setup 10x faster than before
Please add these feature in zoxide it will more fast because zoxide written in rust.

```
zpath_add() {
  for base in "$@"; do
    find "$base" -type d -print0
  done | while IFS= read -r -d '' dir; do
    zoxide add "$dir"
    #echo "✅ Added: $dir"
  done
}
# zpath_add ~/.config ~/.garudrecon ~/some/other/path

# Recursive Add all directories path from input directories
~/.config
~/.config/waymore
~/.config/waymore/results
```

```
zpath_remove() {
  zoxide query -l | grep -E "$1" | while read -r dir; do
    zoxide remove "$dir"
    #echo "❌ Removed: $dir"
  done
}
# zpath_remove "~/.config|~/.garudrecon|sourcemapper"

# remove all paths anywhere input matches, e.g. "sourcemapper"
~/.garudrecon/scans/xyz.app/sourcemapper/cc-fe
~/.garudrecon/scans/xdfhfr.com/sourcemapper/gjh-pkj
```

w3irdrobot:
Adds a `--recursive` flag to the `add` and `remove` commands. 

Closes #1087

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -705,6 +705,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28d3b2b1366ec20994f1fd18c3c594f05c5dd4bc44d8bb0c1c632c8d6829481f"
 
 [[package]]
+name = "same-file"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93fc1dc3aaa9bfed95e02e6eadabb4baf7e3078b0bd1b4d7b6b0b68378900502"
+dependencies = [
+ "winapi-util",
+]
+
+[[package]]
 name = "semver"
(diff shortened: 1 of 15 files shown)


ajeetdsouza:
You don't need walkdir when using the remove command - simply remove any path that starts with the prefix provided as input.

PR #1129, opened 2025-11-04
Add nushell completion for __zoxide_z

Juhan280:
Nushell version `0.109.0` (https://github.com/nushell/nushell/pull/15888) added support for completions to replace multiple arguments. This allows us to define a completion which replaces all the keywords with the correct file path upon accepting the path.

~~Also wrapped the whole script into a nushell module to prevent the completion command from polluting the command namespace. (That's why (the indentation) the diff is so messy, sorry)~~

---

Minimum nushell version required: `0.110.0`


ysthakur (commented):
Random Nushell user here. Thanks for this! I have some minor comments

templates/nushell.txt
```diff
@@ -3,97 +3,127 @@
...
 
-{%- endif %}
+  {%- endif %}
 
-{{ section }}
-# When using zoxide with --no-cmd, alias these internal functions as desired.
-#
+  {{ section }}
+  # Completion for __zoxide_z
+  #
```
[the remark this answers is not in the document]
Juhan280: To match the formating of other section headers in the file
```nu
  {{ section }} 
   # Hook configuration for zoxide. 
   #
```
```nu
  {{ section }} 
   # When using zoxide with --no-cmd, alias these internal functions as desired. 
   #
```

templates/nushell.txt
```diff
@@ -3,97 +3,127 @@
...
-    [ '-' ] => {'-'},
-    [ $arg ] if ($arg | path expand | path type) == 'dir' => {$arg}
-    _ => {
-      ^zoxide query --exclude $env.PWD -- ...$rest | str trim -r -c "\n"
+    let completions = ^zoxide query --exclude $env.PWD --list -- ...$ast.content
+    | lines | first 10 | wrap value
+    | insert span { start: ($ast | first).span.start, end: ($ast | last).span.end }
+
+    {
+      options: {
```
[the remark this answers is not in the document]
Juhan280: Okay

Juhan280:
Minimum nushell version required to make zoxide work without completion: `0.106.0`
To make completion script work properly: `0.109.0`

ysthakur:
<s>@Juhan280 This works great, but if I have multiple arguments, the way the suggestions get replaced isn't super helpful. For example, if I type `z / m`, I might see suggestions for `"~/reedline-main"`, among other paths. If I then choose `"~/reedline-main"`, my buffer will be `z / "~/reedline-main"` rather than `z "~/reedline-main"`.

This is really a limitation of Nushell, and we need to allow completers to replace arbitrary spans there. [My closed PR](https://github.com/ajeetdsouza/zoxide/pull/1064) handled this, but in a very hacky way.</s>

edit: the above problem was just me being stupid, this code works perfectly fine

---

@ajeetdsouza Could we get this PR merged? Despite the problem mentioned above, I think it's good enough for completing single arguments (we can make a new PR in the future once Nushell's completion story improves). If it makes any difference, I'm one of the Nushell maintainers and can confirm that the code in this PR works.

Juhan280:
@ysthakur, what nushell version have you tried? You need `0.109.0` to replace arbitary span as I mentioned above.

ysthakur:
@Juhan280 I'm really sorry, I was just being really dumb above. I completely forgot that we already added the ability to specify spans in custom completers (I'm the one who implemented it).

I have 0.110 installed locally, but because of the `shell.nix` in the Zoxide repo, direnv was automatically switching to Nushell 0.104 in the Zoxide directory. I tried again with Nushell 0.110 and it works just fine. Sorry for the bother.

fdncred:
@ajeetdsouza when you get a chance can you look at this one please and let us know your thoughts?

ajeetdsouza:
Hey! If the Nushell completions API is reasonably stable, I'd be happy to merge zoxide completions for it. However, we need it to match the experience on other shells (bash/zsh/fish). The current behavior is roughly:

- `z foo<Tab>` - local directory completions, since `z` is also a `cd` command
- `z foo<Space><Tab>` - starts up `zoxide query -i` for fuzzy selection, which replaces all the arguments with the result

This has been somewhat fiddly on the other 3 shells, I've just fixed it on the latest `main`. By contrast, I ran `zoxide query -i` inside a Nushell completions function as a test and it was *really* smooth - I think creating a polished Space-Tab experience should be quite doable.

Juhan280:
Hey! Thanks for testing it out.
I'd love your input on one specific UX trade-off with the `Space + Tab` implementation.
Currently, to make this work natively inside Nushell, this PR uses `zoxide query -l` and feeds it to Nushell's completion engine. To preserve `zoxide`'s frecency ranking, I explicitly **disabled** Nushell's native fuzzy finder (otherwise, Nushell overrides the sort order).
Because of this, the behavior is slightly different from Bash/Zsh:
- **Bash/Zsh (via `fzf`):** `z foo <Space><Tab>` opens `fzf`. Typing `bar` does a secondary fuzzy search on the displayed list.
- **This PR (Native Nushell):** `z foo <Space><Tab>` opens the native menu. Typing `bar` just appends to the command, meaning `z foo <Tab>bar<Enter>` is evaluated exactly the same as `z foo bar<Enter>` by `zoxide`. It lacks that secondary visual fuzzy-filtering layer.


Are you comfortable with this native approach, since it relies entirely on `zoxide`'s own ranking algorithm? Or would you prefer I switch this to spawn `zoxide query -i` so the `fzf` interactive experience is strictly 1:1 with the other shells?

Trying to use `zoxide query -i` for the completion would look something like this
```nushell
  def "nu-complete __zoxide_z" [context: string] {
    let ast = ast --flatten $context | skip 1 # skip the command name

    # If the user has typed a space after the first argument, use the custom
    # completer. If not,  use the built-in directory completion.
    if ($ast | is-empty) { return null }
    if $ast.0.span.end >= ($context | str length) { return null }

    ^zoxide query --exclude $env.PWD --interactive -- ...$ast.content
    # this if expression is optional
    | if $in starts-with $"($nu.home-dir)(char psep)" { str replace $nu.home-dir "~" } else {}
    | wrap display_override
    | insert value { get display_override | debug --raw-value }
    | insert span { start: ($ast | first).span.start, end: ($ast | last).span.end }
    | [$in]
  }
```

ajeetdsouza:
I think the fzf-based sorting would be a better fit for zoxide - the UI works really well for fuzzy search and quick navigation, and even shows you directory contents. Nushell's completion system is honestly pretty great, but for this particular job, I would prefer a specialized directory picker.

Juhan280:
So, I pushed a commit that does that. Let me know if its working how you imagined it to be.

I would like to dog food this for a couple of days before going with a merge, if you are okay with this.

ajeetdsouza:
Looks really good! A couple of minor pointers - some of these may not be fixable on Nushell and it's completely fine if that is the case, just let me know:

- `z foo<Tab>` currently shows both files and directories as completions. Can we make this directories-only?
- On other shells, we found a mechanism to accept the line when `fzf` closes, so you don't have to hit `<Return>` twice. Is this possible here?
- `z foo<Space><Tab>` and then `<Esc>` to close the fuzzy picker throws a `NO RECORDS FOUND` below the current prompt. Can it be removed?
- When returning a value from `<Space><Tab>`, we _always_ quote the result, even if it has no spaces in it (e.g. `z "~/Documents"`). Some shells have primitives to only quote a string if it needs quoting - does this exist in Nushell?
- Sometimes, selecting a result from `fzf` causes the screen to clear (the equivalent of hitting `Ctrl-L`). I think this is probably a bug in Nushell, although I haven't figured out how to consistently reproduce it yet.
- We also need to update the minimum supported Nushell version in `man/man1/zoxide-init.1`.

Also, love the substitution of the homedir with `~`! While this is not consistent with the behavior on other shells, that's something that needs to be fixed in the other shells, not here.

Juhan280:
Hi, thanks for your feedback

> * `z foo<Tab>` currently shows both files and directories as completions. Can we make this directories-only?

It was a nushell bug, should be fixed in next release (https://github.com/nushell/nushell/issues/17856)


> * On other shells, we found a mechanism to accept the line when `fzf` closes, so you don't have to hit `<Return>` twice. Is this possible here?

I don't think it is possible for custom completion to run the whole commandline automatically. That can be achieved with custom keybinds and menus. I will look into that if its actally possible for custom completion.

> * `z foo<Space><Tab>` and then `<Esc>` to close the fuzzy picker throws a `NO RECORDS FOUND` below the current prompt. Can it be removed?

I will look into it

> * When returning a value from `<Space><Tab>`, we _always_ quote the result, even if it has no spaces in it (e.g. `z "~/Documents"`). Some shells have primitives to only quote a string if it needs quoting - does this exist in Nushell?

No, nushell doesnt provide a builtin way for doing this.

> * Sometimes, selecting a result from `fzf` causes the screen to clear (the equivalent of hitting `Ctrl-L`). I think this is probably a bug in Nushell, although I haven't figured out how to consistently reproduce it yet.

I am not sure when/why that happens. I havn't noticed anything.

> * We also need to update the minimum supported Nushell version in `man/man1/zoxide-init.1`.

Okay


> Also, love the substitution of the homedir with `~`! While this is not consistent with the behavior on other shells, that's something that needs to be fixed in the other shells, not here.

Thanks

Juhan280:
> * On other shells, we found a mechanism to accept the line when `fzf` closes, so you don't have to hit `<Return>` twice. Is this possible here?

Not possible. Nushell requires you to manually press enter after selecting a completion.

> * `z foo<Space><Tab>` and then `<Esc>` to close the fuzzy picker throws a `NO RECORDS FOUND` below the current prompt. Can it be removed?

So, from what I have found from my testing, you cannot abort a completion from within the completion function. So, you would would need to abort twice; once for fzf and second time for nushell completion.

> * Sometimes, selecting a result from `fzf` causes the screen to clear (the equivalent of hitting `Ctrl-L`). I think this is probably a bug in Nushell, although I haven't figured out how to consistently reproduce it yet.

I could reproduce it partially, it didn't clear the screen like `Ctrl-L` but instead moved the prompt "after" the fzf menu (which is empty at that moment) when i abort the fzf completion

templates/nushell.txt
```diff
@@ -1,101 +1,123 @@
-{%- let section = "# =============================================================================\n#" -%}
+{%- let section = "# =============================================================================" -%}
 {%- let not_configured = "# -- not configured --" -%}
 
 # Code generated by zoxide. DO NOT EDIT.
 
-{{ section }}
-# Hook configuration for zoxide.
-#
+module zoxide_integration {
```
ajeetdsouza: The module change seems unrelated, can we do that in a separate PR?

templates/nushell.txt
```diff
@@ -1,101 +1,123 @@
...
+  # From version 0.110.0, $nu.home-path has been renamed to $nu.home-dir
+  const homedir = if $nu.home-dir? != null { $nu.home-dir } else { $nu.home-path }
```
ajeetdsouza: Given that we only support Nushell v0.106.0+, do we still need this comment / condition? We should just be able to use `$nu.home-dir` inline - we likely don't need this constant at all.

README.md
```diff
@@ -268,7 +268,8 @@ zoxide can be installed in 4 easy steps:
    > ```
    >
    > **Note:**
-   > zoxide only supports Nushell v0.89.0+.
+   > zoxide requires Nushell v0.106.0+. A newer version is recommended for
+   > completions to work properly.
```
ajeetdsouza:
```suggestion
   > zoxide only supports Nushell v0.106.0+.
```

We typically don't support older versions of Nushell, since Nushell makes breaking changes pretty often.

man/man1/zoxide-init.1
```diff
@@ -45,7 +45,8 @@ Now, add this to the \fBend\fR of your config file (find it by running
     \fBsource ~/.zoxide.nu\fR
 .fi
 .sp
-Note: zoxide only supports Nushell v0.89.0+.
+Note: zoxide requires Nushell v0.106.0 or later. A more recent version is
+recommended for completions to function correctly.
```
ajeetdsouza:
```suggestion
Note: zoxide only supports Nushell v0.106.0+.
```

templates/nushell.txt
```diff
@@ -1,101 +1,123 @@
...
-# Jump to a directory using only keywords.
-export def --env --wrapped __zoxide_z [...rest: directory] {
-  let path = match $rest {
-    [] => {'~'},
-    [ '-' ] => {'-'},
-    [ $arg ] if ($arg | path expand | path type) == 'dir' => {$arg}
-    _ => {
-      ^zoxide query --exclude $env.PWD -- ...$rest | str trim -r -c "\n"
+    ^zoxide query --exclude $env.PWD --interactive -- ...$ast.content
+    | str trim --right --char "\n"
```
ajeetdsouza: This line doesn't seem necessary.

templates/nushell.txt
```diff
@@ -1,101 +1,123 @@
...
+    | if $in starts-with $"($homedir)(char psep)" { str replace $homedir "~" } else {}
+    | wrap display_override
+    | insert value { get display_override | debug --raw-value }
+    | insert span { start: ($ast | first).span.start, end: ($ast | last).span.end }
+    | [$in]
```
ajeetdsouza: This might be easier to read:


```suggestion
    | [{
      display_override: $in,
      value: ($in | debug --raw-value),
      span: {start: ($ast | first).span.start, end: ($ast | last).span.end},
    }]
```

templates/nushell.txt
Juhan280: `0.110.0` is newer than `0.106.0`, so yes we still need it

templates/nushell.txt
Juhan280: I added this because otherwise the new completion function will pollute global scope when `source`d directly. Its the idiomatic way of handling scripts that are also modules in nushell. I can make a separate PR if you still insist

templates/nushell.txt
ajeetdsouza: My bad, I misread that line. Let's just support Nushell v0.110.0+ and remove this condition in that case.

templates/nushell.txt
Juhan280: I added because the `__zoxide_z` had it

Juhan280:
Btw, if we push our minimum nushell version to 0.112.1, we get access to call the builtin `cd` using [`%cd`](https://www.nushell.sh/blog/2026-04-11-nushell_v0_112_1.html#no-more-shadows-call-internals-with), so even if the user has overridden his/her `cd`, zoxide will still function. This might be out of scope of this PR though.

templates/nushell.txt
ajeetdsouza: > I added this because otherwise the new completion function will pollute global scope when `source`d directly.

Tbh, I don't think we should make this a module at all. The output of `zoxide init nushell` is already a valid Nushell module, users can just `use ~/.zoxide.nu *` instead of `source ~/.zoxide.nu`.

templates/nushell.txt
ajeetdsouza: Ah, I think this is an artifact from when Nushell didn't strip trailing newlines correctly. Let's just leave it in for now.

ajeetdsouza:
> Btw, if we push our minimum nushell version to 0.112.1, we get access to call the builtin `cd` using [`%cd`](https://www.nushell.sh/blog/2026-04-11-nushell_v0_112_1.html#no-more-shadows-call-internals-with), so even if the user has overridden his/her `cd`, zoxide will still function.

This is a good change, I don't mind doing this in the current PR or a new one, whichever you prefer.

---

Also, I discovered a bug with the current completions.

If the user types in more than one argument `z foo bar<Tab>`, fzf gets triggered. However, after selecting a path from fzf, nushell shows `NO RECORDS FOUND` instead of doing substitution. It works if you put a space `z foo bar<Space><Tab>`.

---

> I would like to dog food this for a couple of days before going with a merge, if you are okay with this.

Once the remaining comments are addressed, I think this should be ready. Let me know once you're satisfied with the PR, and I can merge it.

templates/nushell.txt
Juhan280: It matters because nushell has autoload directory and many users prefer to use that instead of spreading config across two files (env.nu and config.nu):
```nushell
zoxide init nushell | save -f $"($nu.user-autoload-dirs.0)/zoxide.nu"
```

Juhan280:
> Also, I discovered a bug with the current completions.
> 
> If the user types in more than one argument `z foo bar<Tab>`, fzf gets triggered. However, after selecting a path from fzf, nushell shows `NO RECORDS FOUND` instead of doing substitution. It works if you put a space `z foo bar<Space><Tab>`.

I cannot reproduce this. (zoxide 0.9.3, also tried building zoxide from this PR; nushell 0.114.2-git-e3dcf2d9)

https://github.com/user-attachments/assets/f757311e-b1f3-4b37-a1d8-8d3f818569a9

---
> > Btw, if we push our minimum nushell version to 0.112.1, we get access to call the builtin `cd` using [`%cd`](https://www.nushell.sh/blog/2026-04-11-nushell_v0_112_1.html#no-more-shadows-call-internals-with), so even if the user has overridden his/her `cd`, zoxide will still function.
> 
> This is a good change, I don't mind doing this in the current PR or a new one, whichever you prefer.
> 

I would prefer a separate PR.

templates/nushell.txt
ajeetdsouza: That seems like a discussion for a separate issue / PR - let's not mix it up with this one. Can we remove the modules change for now?

Juhan280: alright

Juhan280:
We need to hold on for some time, it seems like upstream is experimenting on disallowing interactive process inside the completion function. If that happens, we need to either use the builtin completion menus, or define a custom keybind
- https://github.com/nushell/nushell/issues/18771

Issue #1134, opened 2025-11-10
zoxide messes with starship's [status.pipestatus]


playbahn:
[`starship.rs`](https://github.com/starship/starship) has a `[status]` module, which has a `pipestatus` key. Setting up shell integration for zoxide messes it up.

<img width="363" height="251" alt="Image" src="https://github.com/user-attachments/assets/7cd45f92-ddb3-4120-bce9-2774115ea1cd" />

```console
$ bat ~/.bashrc -r125: --plain
eval "$(starship init bash)"
# ----------------------------------------------------------------------- zoxide
# fucks up starship's [status.pipestatus]
# eval "$(zoxide init bash)"
# ------------------------------------------------------------------------- tmux
if [[ $TERM_PROGRAM != @(tmux|vscode|zed) && "$DISPLAY" && -x "$(command -v tmux)" ]]; then
    if [[ "$(tmux list-sessions -F '69' -f '#{==:#{session_attached},0}' 2> /dev/null)" ]]; then
        tmux attach-session
    else
        tmux new-session
    fi
fi
```

ajeetdsouza:
Fixed, thanks for reporting this!

PR #1136, opened 2025-11-12
Make zoxide compatible with starship.rs prompt

playbahn:


playbahn:
#1134

ajeetdsouza:
This PR modifies the `__zoxide_doctor` function - it would not fix the issue you're mentioning. `doctor` currently identifies that zoxide has been incorrectly initialized before Starship - that behavior should not be changed.

playbahn:
> This PR modifies the `__zoxide_doctor` function - it would not fix the issue you're mentioning. `doctor` currently identifies that zoxide has been incorrectly initialized before Starship - that behavior should not be changed.

Apologies for the poorly worded PR title and commit msg. `zoxide` actually works just fine with `starship` (seems so). But `__zoxide_doctor` is run every time `z` is run, and so it checks `PROMPT_COMMAND`. The problem is how starship modifies `PROMPT_COMMAND`

```console
$ starship init bash --print-full-init | bat -r125:136 --plain -lbash
```
```bash
    # Finally, prepare the precmd function and set up the start time. We will avoid to
    # add multiple instances of the starship function and keep other user functions if any.
    if [[ -z "${PROMPT_COMMAND-}" ]]; then
        PROMPT_COMMAND="starship_precmd"
    elif [[ "$PROMPT_COMMAND" != *"starship_precmd"* ]]; then
        # Appending to PROMPT_COMMAND breaks exit status ($?) checking.
        # Prepending to PROMPT_COMMAND breaks "command duration" module.
        # So, we are preserving the existing PROMPT_COMMAND
        # which will be executed later in the starship_precmd function
        STARSHIP_PROMPT_COMMAND="$PROMPT_COMMAND"
        PROMPT_COMMAND="starship_precmd"
    fi
```

So if anything that modifies `PROMPT_COMMAND` (zoxide integration in this case) is put after starship integration, it ends up breaking starship's functionality. Starship however does this:

```bash
STARSHIP_PROMPT_COMMAND="$PROMPT_COMMAND"
PROMPT_COMMAND="starship_precmd"
``` 

So now if zoxide is put before starship, starship works but `__zoxide_hook` ends up being inside `STARSHIP_PROMPT_COMMAND`. At which point `doctor` can't find `__zoxide_hook` in `PROMPT_COMMAND`, showing error messages. The VSC compatibility line essentially does the same thing (*"`PROMPT_COMMMAND` might have been changed, check if this other env var exists"*) :

```bash
[[ ${__vsc_original_prompt_command[@]:-} == *'__zoxide_hook'* ]] && return 0
```

Likewise, starship compatibilty check would be:

```bash
[[ ${STARSHIP_PROMPT_COMMAND[@]:-} == *'__zoxide_hook'* ]] && return 0
```

---

> Appending to PROMPT_COMMAND breaks exit status ($?) checking

exit status (the `[status]` module) contains `pipestatus`:

```toml
[status]
format = '[]($style)[$status]($style inverted)([ $signal_name $signal_number]($style inverted))[]($style) '
style = 'bold #ff0040'
disabled = false
map_symbol = false
pipestatus = true
pipestatus_format = '[]($style)[$pipestatus]($style inverted)[]($style)([]($style)[$signal_name\($signal_number\)( $maybe_int )]($style inverted)[]($style)) '
pipestatus_segment_format = '[$status]($style inverted)'
pipestatus_separator = '  '
```

ajeetdsouza:
zoxide should always be initialized at the end of your bashrc. I know this is suboptimal and potentially restrictive, but the reason for this is that custom configs and other plugins often silently break zoxide by incorrectly overriding `$PROMPT_COMMAND`.

As a maintainer, debugging this via GH issues is quite time-consuming, and I can't debug every config / plugin that breaks zoxide. Sometimes users don't realize that their hook is broken, and assume zoxide is not working well. Instead, my solution was to require zoxide to be initialized last (enforced by `__zoxide_doctor`), and ensure that zoxide correctly handles the hooks of all other plugins. In this case, when initialized last, zoxide breaks the PIPESTATUS segment of Starship, which should be treated as an urgent bug in zoxide (fixed by https://github.com/ajeetdsouza/zoxide/pull/1141).

`__zoxide_doctor` should continue to warn users when zoxide is overridden by Starship, because that means it is not being initialized last. For example, consider a use case like this:

```bash
# ~/.bashrc
eval "$(zoxide init bash)"
eval "$(starship init bash)" # correctly handles zoxide

### user forgets about Starship
PROMPT_COMMAND="echo '>> '"
# now zoxide is broken, and __zoxide_doctor won't report it since it's in STARSHIP_PROMPT_COMMAND
```

The reason for the `__vsc_original_prompt_command` exception is that VS Code initalizes the prompt after `.bashrc` has already run, which means there is no way to initialize zoxide after that.

Hope that makes sense!

playbahn:
You're right. That does make sense. Thanks for bearing with me :)

PR #1139, opened 2025-11-17
Feat/fuzzy search

### LINKED ISSUE #638: Fuzzy search is not matching strings as desired
First of all sorry, was not sure how to formulate issue title correctly, gonna try to explain issue bellow.

Came to zoxide after years of using autojump.

Simple example of issue:

Searching for `docs` is not matching with `~/Documents` in either `z` or `zi`. Have to type exactly `doc`.

The issue did not appear in `autojump`, and even `fzf` outputs result """correctly""".

### LINKED ISSUE #1137: Try fuzzy search when no match found
Hello,

I'm using zoxide for a while, what bothers me a lot is that I always have typos when typing, for example, `applicaiton` -> `application`. In this case zoxide just returns `no match found`, but I think the user experience could be better if zoxide could use fzf to do fuzzy search when there's `no match found`, that would help A LOT for users who is really a typo producer LOL. 

In the last I want to say thanks for creating zoxide, it's really great.

chojs23:
Add typo-tolerant fuzzy subsequence matching within path components so partial inputs like docs ,dcmts or doucments resolve to ~/Documents without breaking existing slash semantics.

Closes #638 , Closes #1137

### DIFF
--- a/src/db/stream.rs
+++ b/src/db/stream.rs
@@ -85,19 +85,21 @@ impl<'a> Stream<'a> {
 
         let path = util::to_lowercase(path);
         let mut path = path.as_str();
-        match path.rfind(keywords_last) {
-            Some(idx) => {
-                if path[idx + keywords_last.len()..].contains(path::is_separator) {
+
+        let (idx, end) = match rfind_component_match(path, keywords_last) {
+            Some((idx, end)) => {
+                if path[end..].contains(path::is_separator) {
                     return false;
                 }
-                path = &path[..idx];
+                (idx, end)
             }
             None => return false,
-        }
+        };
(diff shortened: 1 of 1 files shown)


stickynotememo:
Any update on this? Would be a useful feature.

ljbred08:
Please add this! I just came across this and this was the first thing I expected it to do, advertising as a "smarter cd".

PatrickF1:
+1 was a bit surprised it doesn't handle fuzzy matching

chojs23:
@ajeetdsouza Could you check this?

Hund:
I would love to see this feature! This is something that I truly miss coming from autojump.

PR #1143, opened 2025-11-18
fix: prefer exact keyword match over score match

### LINKED ISSUE #1130: prefer exact match on last folder name
I have folders `~/opt/linux` and `~/opt/linux-stable`. Currently, `~/opt/linux-stable` is ranked higher, so when I type `z linux` I go to `~/opt/linux-stable`. I believe that if the query is an exact match of the name of the folder, it should go to that folder.

dracarys18:
Fixes #1130

I had this issue when I am trying to cd to `linux` it would always choose `linux-mma` because it was frequently accessed and accessed recently. So add an additional sorting over the score sorting which is based on exact keyword matching. This would always prefer exact matches over the score


mghesh-yseop:
Please @ajeetdsouza could you take a look here ? I have the same issue, which is really annoying day to day.
I think having a configuration option could be used to activate this feature though, as this can break the default behavior.

xatuke:
i agree, this is a huge issue for me as well. especially since my projects are named tmp-[project name] and I have a tmp folder as well

mghesh-yseop:
@ajeetdsouza I'd like to know if it is planned or not to get this behaviour?

For the moment, I don't have another choice but to stop using `zoxide`, as I always have the wrong destination for many directories...

I'll stay updated for any news, hoping to have this one day.

xamcost:
Doubling on this, it would be a really nice feature.

dracarys18:
I will add this under a config if that's preferable.

mghesh-yseop:
If you could add the associated documentation to this feature, it would be very nice :smile:

robin-vagner:
Bumping this, it would be a really great feature!

PR #1149, opened 2025-11-18
Support symlinks on Windows

### LINKED ISSUE #1147: Windows can have symlinks, should still respect `_ZO_RESOLVE_SYMLINKS`
The logic here ignores the value of `_ZO_RESOLVE_SYMLINKS` on Windows, but Windows can have symlinks.
https://github.com/ajeetdsouza/zoxide/blob/ecf72c9cfdc0c8f04ac1e2953bb3703ea6805e05/templates/posix.txt#L10-L19

Here is a reproducible example from my busybox.exe (ash) shell:

```bash
# Create directory and symlink to it
$ cd G:\repro
$ mkdir directory
$ ln -s directory symlink

$ cd symlink
$ pwd -L
G:/repro/symlink
$ pwd -P
G:/repro/directory

# cygpath is not available on $PATH and is not necessary in busybox.
# But for completeness, just showing that it understands the symlink, too.
$ 'C:\program files\git\usr\bin\cygpath.exe' -w $(pwd -L)
G:\repro\symlink
$ 'C:\program files\git\usr\bin\cygpath.exe' -w $(pwd -P)
G:\repro\directory
```

### LINKED ISSUE #1148: Posix shells on Windows might not have, nor require, cygpath
`busybox.exe sh` is the `ash` shell on Windows. It uses native windows paths -- `G:\foo\bar` or `G:/foo/bar` -- so it doesn't need nor provide cygpath.

Is it reasonable for `zoxide init` to only call `cygpath` conditionally, if it's found, skipping it otherwise?

```bash
# In busybox.exe sh, we already get native windows paths, albeit with forward slashes
$ pwd -P
G:/repro
$ cd 'G:\repro' # <-- It understands backslashes, too
```

I think this implementation avoids any performance hit by checking for cygpath only once on startup, baking the decision into the `__zoxide_pwd` implementation:

```
if \command -v cygpath >/dev/null
then
  __zoxide_pwd() {
    \command cygpath -w "$(\command pwd -P)"
  }
else
  __zoxide_pwd() {
    \command pwd -P
  }
fi
```

cspotcode:
Respects `_ZO_RESOLVE_SYMLINKS` the same on Windows as it already does on Linux. Fixes #1147 

Checks for the presence of `cygpath` and skips if it's not present. This supports non-Cygwin shells like `busybox.exe` that do not require nor provide `cygpath`, because they already use valid, native Windows paths.  Fixes #1148

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- POSIX: support for non-Cygwin Windows environments (e.g. Busybox).
+
+### Fixed
+
+- Bash/Fish/POSIX/Zsh: resolve symlinks on Windows.
+
 ## [0.9.9] - 2026-01-31
 
 ### Added


ajeetdsouza:
Thanks for this change. Given that Busybox Bash is aliased to Ash (which is a POSIX shell), I think it make more sense to make this change only to `posix.txt`. Busybox users should use `zoxide init posix`.

PR #1165, opened 2026-01-29
fish: clear existing completions before defining aliases

simono:
Fish ships with built-in completions for common commands like `j` (for autojump): https://github.com/fish-shell/fish-shell/blob/master/share/completions/j.fish

When using `zoxide init --cmd j fish`, these completions override zoxide's and break directory completion.

This PR clears existing completions before defining aliases, so zoxide's completions take effect.


ajeetdsouza:
Thanks!

Issue #1166, opened 2026-01-30
Relative paths involving symlinks not detected as directories


br0ns:
This concerns at least `zsh`, I haven't tried other shells.

Steps to reproduce:

```
mkdir -p foo/bar
ln -s foo/bar baz
cd baz
cd ../foo
```

The last command fails with `zoxide: no match found`.

The problem lies the second `if`-statement in `templates/zsh.txt` which uses `test -d` to check if `../foo` is a directory, which dereferences symlinks _before_ expanding `..`.  Therefore `../foo` is determined to not be a directory and Zoxide is queried instead of calling the builtin `cd`.  Note that `cd` works as expected in this scenario.

ajeetdsouza:
Fixed in https://github.com/ajeetdsouza/zoxide/commit/63ef5518e2c9805e3434cc5bb607ae0cc9631e21. Thanks for reporting this!

PR #1167, opened 2026-01-30
fix issue #1166

br0ns:
This PR fixes issue #1166 by using `realpath -L` to expand `..` before symlinks when detecting whether a target path is a directory or not.


ajeetdsouza:
Hey, thanks for the PR! This is a good way to do it, but I found a way that would fix this and also CDPATH in one go: https://github.com/ajeetdsouza/zoxide/commit/63ef5518e2c9805e3434cc5bb607ae0cc9631e21

Issue #1174, opened 2026-02-04
[FEATURE REQUEST] using unique child directory to pinpoint multiple results


huttli:
I have very similar directory names, e.g. `abc-api`, `abc` and `abc-ui`. It's possible to use the unique part of the directory name to target the `abc-api` and `abc-ui` directory, but not for `abc` directory. I get almost half of the time miss on the `abc` where zoxide just goes to `abc-api` or `abc-ui` based on most recently used one.

For my use case, I do have very unique child directory names inside `abc`. I'm wondering if it's possible to have a feature like this `z abc/uniq_child_dir/..`.

I did do a simple search in both issues and pull requests and didn't find a similar feature.

PR #1175, opened 2026-02-05
Feat: use uniq child to target parent directory

huttli:
This is an LLM generated change to address https://github.com/ajeetdsouza/zoxide/issues/1174. It seems to be working on my local test for my use cases.


ajeetdsouza:
Sorry, this is not a planned feature.

PR #1181, opened 2026-02-18
feat(import): add support for importing from Jump

vegerot:
Add [Jump](https://github.com/gsamokovarov/jump) as a new source for the `zoxide import` command. The implementation includes a parser for Jump's data format and updates to all shell
completions (bash, fish, zsh, nu, fig) to include "jump" as an option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1003,6 +1003,7 @@ dependencies = [
  "rstest",
  "rstest_reuse",
  "serde",
+ "serde_json",
  "tempfile",
  "which",
 ]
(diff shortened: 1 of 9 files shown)


ajeetdsouza:
Use serde_json for this - after LTO, it doesn't increase the binary size by much.

vegerot:
@ajeetdsouza done!  btw I can shave off another 60 lines to this patch if I replace my handrolled timestamp parser with `chrono`.  The one I wrote works fine, but lmk

ajeetdsouza:
I think a `SystemTIme` should work fine here - no need of importing `chrono`.

PR #1215, opened 2026-04-06
feat: shorten /home/user to ~ in fzf

### LINKED ISSUE #1162: Have `zi` command shorten `/home/$USER/…` paths to `~/…`
In fzf selector, I would like for dirs that start with `$HOME` in path to just be prefixed with `~` instead of full `/home/$USER/` as most of my frequent paths are in my home dir and I don't need to see my home dir's name.

abghim:
This closes #1162.

## Before
Using fzf in zoxide (through `zi` or `zoxide query -i`) listed all directories as `/home/user/...` or `/Users/user/...`.

## After
The fuzzy finder shortens the prefix to `~`. Other features remain intact & functional as before.
image

## Notes
This commit impacts just the fzf component, used in `zi`. The fuzzy finder will no longer display absolute paths as /home/user or /Users/user (macOS). These prefixes will be shortened to '~'. This affects just fzf visuals; that is, the final output of fzf still displays the full absolute path. `ls` preview within fzf and final `zi` command all work as expected. Works correctly on macOS & Linux.

This is arguably a big UI change, I totally understand if the devs are reluctant to merge this.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -60,3 +60,7 @@ pub fn maxage() -> Result<Rank> {
 pub fn resolve_symlinks() -> bool {
     env::var_os("_ZO_RESOLVE_SYMLINKS").is_some_and(|var| var == "1")
 }
+
+pub fn get_home_dir() -> OsString {
+    env::var_os("HOME").unwrap_or("".into())
+}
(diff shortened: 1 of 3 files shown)


abghim:
@ajeetdsouza Perhaps this feature doesn't align closely with the intended direction of the project. This is a big user-facing change; you can close this if you want :)

Issue #331, opened 2022-01-09
could not get drive letter: \\a.network-drive\a-folder


ajeetdsouza:
Using zsh in MSYS2 on Windows. When I cd into a network drive like `//a.network-drive/a-folder`, zoxide will give me:

```
zoxide: could not get drive letter: \\a.network-drive\a-folder
```

PR #1228, opened 2026-05-04
fix: handle VerbatimUNC paths on Windows (issue #331)

mitre88:
Fixes issue #331 - handles UNC paths like \\a.network-drive/a-folder on Windows by adding support for Prefix::VerbatimUNC in the resolve_path function.


mitre88:
Closing this to keep my open PR queue focused on contributions that are currently reviewable and likely to proceed. This one has a blocking signal (failing-checks) and would need a deeper rework before it is worth maintainer time. Thanks for the review/context.

Issue #1116, opened 2025-09-16
alternative/additional operation mode the simply completes the most recent directories


calestyo:
Hey.

Currently, AFAIU, `zoxide` stores all the dirs one changed to (via it) in some db and then selects/completes based on that.

It would be nice if one could additionally/alternatively (and make that configurable) have a mode where it presents the most recent current directories in order.
So roughly, e.g. if I do:
```
~$ z tmp
~/tmp$ z /etc
/etc$ z /tmp
/tmp$ z <see below>
```
it would use:
```
1: /etc
2: ~/tmp
3: ~
```

I'd kinda envision two ways to access this:
```
z -<n>
```
where `<n>` is number and `-1` is the same as `-` (i.e. the previous directory), but `-2` is the one before the previous and so on. Here the user would obviously need to know which one the -2nd was.<br>
Or alternatively:
```
z -<TAB>
```
which would cause `fzf` to pop up and simply display the list from the previous, to the one before that and so on.

The whole idea is a simpler (but of course also more limited) but much faster mode than the current. If I know that I want to got to the nth previous dir, then typing `-<n>` is faster.<br>
Also, sometimes I know that I've been to the right directory, a few `cd`s before but don't really remember which one it was (and also don't have any good search string for it). The current mode doesn't help me that much now, but the one I propose here would give me simply the history of my most recent directories and I could look up, which I wanted.

For that reason, if doing the `-<TAB>` completion and getting the list in `fzf`, it might probably be useful, if `fzf` would then **only** show te most recent ones and no other possible candidates (or at least it should be configurable whether or not others should be included).<br>
And as already said above, I think it would be good if one could configure which of the two modes one wants (both or either of them).

On top of that (as if it wouldn't already be enough 😇) ... it would be nice if one could configure whether the proposed mode works per shell or has a shared list of most recent dirs, over all shells.<br>
Would be nice if one could set a global default for this (e.g. use a per-shell list per default), but could perhaps even change during the lifetime of a shell via some `z --option` from one to the other.<br>
If per shell (which limits the list obviously to the lifetime of the shell) it would probably make sense to store the list in some ephemeral location (like `/run/user/`) - why causing disk writes if it's anyway only temporary? Storing the list in a shell variable would probably be tricky, as one would need a separator char (but then one needs a quoting system for that), unless of course one has e.g. arrays like with bash.

[...]

PR #1233, opened 2026-05-10
feat(db): add opt-in _ZO_RANKING_MODE=recency

### LINKED ISSUE #1231: Most-recently-used mode
Is there a way to configure zoxide such that it's purely recency based? 

e.g. if I type `z opt` I would like to go to the most recently visited directory containing `opt`, not some project I visited a bunch of times 3 months ago (and again just now, bumping its ranking even higher) that I now have to call `zoxide remove` on.

ChrisJr404:
Closes #1231.

Adds an opt-in `_ZO_RANKING_MODE` env var that lets users switch the query sort from the default frecency to plain recency:

- `frecency` (default): existing behaviour, no change.
- `recency`: sort by `last_accessed`, so a directory you just visited always beats a popular one from months ago.

The value is case-insensitive. Unknown values fall back to `frecency`, so a typo can't silently change behaviour.

### Why opt-in

The wiki explicitly calls out that the algorithm is intentionally predictable, and there's an older similar request (#1116) that also asks for an alternative mode. Defaulting to recency would change everyone's results, so I kept this behind an env var. The diff is small (~100 lines) and reversible: deleting `RankingMode::Recency` and the one env read in `config.rs` removes the feature entirely.

### Implementation notes

- `RankingMode` lives in `config.rs` next to the other `_ZO_*` readers.
- `Dir::score` now takes the mode explicitly. This keeps it a pure function with no env reads in the sort hot path, and makes the new tests trivial to write.
- `sort_by_score` and `StreamOptions` both gained a mode parameter; the two command-site callers (`cmd/query.rs`, `cmd/edit.rs`) pick it up via `config::ranking_mode()`.
- `DirDisplay` keeps showing the frecency value in both modes so `zoxide query --score` output stays comparable across modes and downstream consumers don't break.
- On-disk format is unchanged.

### Tests

- 3 new unit tests in `db::dir::tests` cover: frecency prefers high-rank, recency prefers recent, recency is rank-independent.
- Existing 16 tests still pass.
- Manual smoke test with two directories (one popular but stale, one fresh):
  - default mode lists the popular one first
  - `_ZO_RANKING_MODE=recency` lists the fresh one first
  - `_ZO_RANKING_MODE=Recency` (mixed case) and `_ZO_RANKING_MODE=nonsense` (fallback) both behave as documented

### Docs

- README env-var section + new `[algorithm-frecency]` link.
- `man/man1/zoxide.1` ENVIRONMENT block.

Kept this minimal to start. Happy to expand if you'd like a deeper integration, e.g. shell-local recency lists like #1116 sketches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,6 +426,13 @@ Environment variables[^2] can be used for configuration. They must be set before
   - Configures the [aging algorithm][algorithm-aging], which limits the maximum
     number of entries in the database.
   - By default, this is set to 10000.
+- `_ZO_RANKING_MODE`
+  - Selects the ranking function used when sorting query results:
+    - `frecency` (default): combines frequency and recency, as described in
+      [Algorithm][algorithm-frecency].
+    - `recency`: sorts purely by `last_accessed`, so the most recently visited
+      match always wins regardless of how often older directories were used.
+  - The value is case-insensitive. Unrecognized values fall back to `frecency`.
 - `_ZO_RESOLVE_SYMLINKS`
   - When set to 1, `z` will resolve symlinks before adding directories to the
     database.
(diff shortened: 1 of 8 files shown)


ajeetdsouza:
Hi, thanks for putting this together. Unfortunately, this is not a planned feature for zoxide, so I'm closing this.

Bugfixes and minor improvements are generally good candidates for direct PRs, but for new features, please wait for maintainer feedback / approval before implementing them. It helps make sure the work aligns with the direction of the project before you spend time on it.

calestyo:
really a shame :-(

PR #1242, opened 2026-05-22
Accept existing paths in interactive query

### LINKED ISSUE #1239: zi should succeed when the argument is a realpath or can match paths
currently zi fail in such a case and I have to type command again.

It means, I have to *remember* which path is in the db.
That is insane.

puneetdixit200:
Closes #1239.

This updates `zoxide query --interactive` to return a single existing directory path directly before launching fzf. That makes `zi <real path>` behave like a direct jump while keeping the normal fzf flow for non-path queries, multiple keywords, and score output.

I also added an integration test for the direct existing-path case.

Tested:
- `cargo fmt --all -- --check`
- `cargo test --test query --no-default-features`
- `cargo test db::stream::tests::query --no-default-features`
- `cargo test --workspace --no-default-features --no-fail-fast`
- `cargo clippy --all-features --all-targets -- -Dwarnings`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `zoxide query --interactive` now accepts a single existing path directly.
 - PowerShell: use fully qualified names when invoking cmdlets.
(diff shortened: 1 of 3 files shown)


ajeetdsouza:
Sorry, this is not a planned feature - `zi` is explicitly designed to be an interactive selector / search.

PR #1250, opened 2026-06-15
Add feature to batch-rename entries

### LINKED ISSUE #916: Cannot edit paths in the zoxide database
I recently renamed my home folder (e.g. from `/home/old` to `/home/new`). Unfortunately this rendered my precious zoxide database useless.

I'd like to be able to replace occurrences of `/home/old` to `/home/new` in the zoxide database, but I haven't figured out how. 

What I've tried:
- `sed -i s/old/new/g ~/.local/share/zoxide/db.zo`. This breaks the database (understandably).
- `zoxide edit`. Doesn't seem to provide a way to actually edit paths, just scores.
- `zoxide import`. This sounded promising, but there is no `zoxide export`.

Something that allows editing the db as a plaintext file would be amazing for these cases.

alcremieow:
Closes #916.

This adds `zoxide rename --old-name <old_name> --new-name <new_name>`.
This is mainly intended for the use case described in #916 (migrating from `/home/old-user` to `/home/new-user`), as currently there is no good way to edit paths in the zoxide database while keeping their `rank` and `last_accessed` time.

The subcommand replaces all occurences of `old-name` with `new-name`, prints out the changes and asks the user whether they want to apply the changes (y/n).

### DIFF
--- a/contrib/completions/_zoxide
+++ b/contrib/completions/_zoxide
@@ -209,6 +209,18 @@ _arguments "${_arguments_options[@]}" : \
 '*::paths:_files -/' \
 && ret=0
 ;;
+(rename)
+_arguments "${_arguments_options[@]}" : \
+'--old-name=[]:old_name:_files -/' \
+'--new-name=[]:new_name:_files -/' \
+'-f[skip y/n]' \
+'--force[skip y/n]' \
+'-h[Print help]' \
+'--help[Print help]' \
+'-V[Print version]' \
+'--version[Print version]' \
+&& ret=0
+;;
         esac
     ;;
 esac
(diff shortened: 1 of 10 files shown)


Eviber:
I just needed this feature and for what its worth, I can confirm that it worked as expected. Thanks! 😄

HarrisonMc555:
For anyone who wants to batch rename paths today, I created this script (originally commented on #916):

```bash
#!/usr/bin/env bash

old_path="/my/old/path"
new_path="/my/new/path"

zoxide query -a -s -l "$old_path" / |
    while read -r line; do
        score="${line%% *}"
        old_entry="${line#* }"
        new_entry=$(echo "$old_entry" | sed "s#${old_path}#${new_path}#")
        echo "${old_entry} -> ${new_entry}"
        if ! zoxide remove "$old_entry"; then
            >&2 echo "Could not remove '${old_entry}'"
            break
        fi
        if ! zoxide add "$new_entry" -s "${score}"; then
            >&2 echo "Could not add '${new_entry}'"
            break
        fi
    done
```

PR #1256, opened 2026-06-30
fix(zsh): detect relative paths through symlinks as directories

### LINKED ISSUE #1166: Relative paths involving symlinks not detected as directories

salarkhannn:
When the current directory is a symlink, `[[ -d ../foo ]]` resolves the symlink before expanding `..`, so the directory check fails even though `cd ../foo` would succeed using the shell's logical path tracking.

Fix: also match paths containing `..` so they fall through to `__zoxide_cd`, which relies on the shell's native `cd` behavior.

Closes #1166


ajeetdsouza:
Thanks for the PR. However, this would break queries that used `..` as keywords. I found a fix that should work: https://github.com/ajeetdsouza/zoxide/commit/63ef5518e2c9805e3434cc5bb607ae0cc9631e21

PR #1257, opened 2026-07-04
repaint the prompt when exiting without selecting so the command line isn't left in a broken state

developic:
When the interactive picker is canceled without selecting a value, zoxide query --interactive returns no result...... The function had no fallback for this case, so it left the command line in a broken state...


ajeetdsouza:
Hi, what do you mean by "left the command line in a broken state"? Also, it looks like your PR is executing `cd` when the user has canceled the interactive picker, and doing nothing when the user actually picks something.

developic:
> Hi, what do you mean by "left the command line in a broken state"? Also, it looks like your PR is executing `cd` when the user has canceled the interactive picker, and doing nothing when the user actually picks something.

if i exit fzf its do not repaint the command line

ajeetdsouza:
Are you on the latest version of zoxide (0.10.0)?

```bash
zoxide --version
```

If yes, can you upload a GIF / series of screenshots of how that looks? I'm not able to reproduce this on my system.

developic:
> Are you on the latest version of zoxide (0.10.0)?
> 
> ```shell
> zoxide --version
> ```
> 
> If yes, can you upload a GIF / series of screenshots of how that looks? I'm not able to reproduce this on my system.

image
image

the prompt is broken as you  can see when i exit 
```
 ~
❯ zoxide -V
zoxide 0.9.9
```


after fixed:

https://github.com/user-attachments/assets/7c18555d-8fd8-4483-bf7b-a4d62430bdb1

developic:
Sorry for the late response, I was a little busy with my personal life......

developic:
It's not a problem with zoxide. The problem is with the Fish script, which doesn't handle the function's exit correctly...

Issue #1021, opened 2025-03-20
z doesn't register directory in PowerShell


cy-james-lee:
```
~> z C:\users
C:Users>z
~>z users
zoxide: no match found
~>pwsh --version
PowerShell 7.5.0
~>zoxide --version
zoxide 0.9.2
```

Confirmed that `zoxide query -l` is empty.

ajeetdsouza:
Have you:

1. Ensured that zoxide is initialized at the end of your config file
2. cd-ed around normally to populate the database?

cy-james-lee:
@ajeetdsouza Yes, I installed following the instruction and put this `Invoke-Expression (& { (zoxide init powershell | Out-String) })` into my $PROFILE file. Is it possible that the hook isn't working? `function global:__zoxide_hook` must not be doing `zoxide add`.

rijulkap:
I ran into this same issue earlier this week. Do you also use starship by any chance (and have its init called within your powershell script)? Try using Windows' inbuilt powershell instead of pwsh and that should work straight out of the box.

I have also submitted a PR that should fix this issue when using pwsh instead of powershell. [#1023](https://github.com/ajeetdsouza/zoxide/pull/1023)

cy-james-lee:
> I ran into this same issue earlier this week. Do you also use starship by any chance (and have its init called within your powershell script)? Try using Windows' inbuilt powershell instead of pwsh which is what got it working for me.

Thanks for the info. Yes, I'm using Starship. Do we have a workaround for this?

luizxrs:
same problem here, using starship too

darkmusic:
If you use starship, just make sure the starship initialization happens *before* the zoxide initialization.

```powershell
Invoke-Expression (&starship init powershell)
Invoke-Expression (& { (zoxide init powershell | Out-String) })
```

Once you do that, the directory saving/recollection will work correctly in PowerShell.

rijulkap:
> If you use starship, just make sure the starship initialization happens _before_ the zoxide initialization.
> 
> Invoke-Expression (&starship init powershell)
> Invoke-Expression (& { (zoxide init powershell | Out-String) })
> 
> Once you do that, the directory saving/recollection will work correctly in PowerShell.

In the case of Powershell ( the shell inbuilt in Windows, i.e Version 5.1 and below ), this works and everything works as intended.

[...]

Issue #877, opened 2024-08-30
Zoxide is not learning ?


avario-cpu:
Am I doing somethig wrong ?

https://github.com/user-attachments/assets/47747f9c-4762-4990-b81b-d09cb3a3a1ae

avario-cpu:
Had not put
`Invoke-Expression (& { (zoxide init powershell | Out-String) })`
At the bottom of the file.
Since starship requires
`Invoke-Expression (&starship init powershell)`
To be at the bottom as well. 

So priority seems to be zoxide > starship. Starship still works this way

ajeetdsouza:
Yup, zoxide can play nicely with other programs when it's last, but other programs may override zoxide's config when they're initialized after. Unfortunately, there isn't much we can do about that.

rijulkap:
Ran into this issue as well. Can't seem to get zoxide to learn newer directories after installing starship. Tried the recommended fix, but to no avail. My current powershell profile looks as follows:


```
Invoke-Expression (&starship init powershell)

$prompt =""
function Invoke-Starship-PreCommand {
    $current_location = $executionContext.SessionState.Path.CurrentLocation
    if ($current_location.Provider.Name -eq "FileSystem") {
        $ansi_escape = [char]27
        $provider_path = $current_location.ProviderPath -replace "\\", "/"
        $prompt = "$ansi_escape]7;file://${env:COMPUTERNAME}/${provider_path}$ansi_escape\"
    }
    $host.ui.Write($prompt)
}

Invoke-Expression (& { (zoxide init powershell | Out-String) })
```

Removing the middle code block (required by wezterm) does not help. Shuffling the init order or using a different terminal does not help either. The only way to get zoxide to add directories once again is getting rid of the starship init command completely. Any suggestions at all?

ajeetdsouza:
Please create a fresh issue.

PR #1258, opened 2026-07-04
docs: add PowerShell prompt compatibility note to README

### LINKED ISSUE #74: The database is not being created in Windows Powershell
## Steps to reproduce

- Install zoxide: `cargo install zoxide -f`
- Add the hook to `$PROFILE`
- Restart terminal
- Go to some dir with zoxide: `z some_dir`
- Check zoxide dir: `ls C:\Users\UserName\AppData\Local\zoxide` (Error! No such directory!)
- Try to match with zoxide: `z some` (Error! No matches!)
- Check zoxide dir again: `ls C:\Users\UserName\AppData\Local\zoxide` (Ok! Now it exists but there is still no db file) 

## Additional info

I am running the default powershell available on windows (`$PSVersionTable.PSVersion.Major` yields 5, thus the hook defaults to `'prompt'` according to the README powershell snippet)

LingyeNBird:
Add a note to the PowerShell installation section explaining that zoxide must be initialized after prompt theme tools (e.g. oh-my-posh, starship, posh-git) in the user profile.

This limitation is unique to PowerShell because it uses a single `prompt` function that gets overwritten by the last initialization. Other shells like Bash and Zsh use array-based hook mechanisms (`PROMPT_COMMAND` / `precmd_functions`) and are not affected.

Closes: https://github.com/ajeetdsouza/zoxide/issues/74, https://github.com/ajeetdsouza/zoxide/issues/877, https://github.com/ajeetdsouza/zoxide/issues/1021

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,12 @@ zoxide can be installed in 4 easy steps:
    > ```powershell
    > Invoke-Expression (& { (zoxide init powershell | Out-String) })
    > ```
+   >
+   > **Note for PowerShell users:** If you use a prompt theme tool (such as
+   > oh-my-posh, starship, or posh-git), make sure zoxide is initialized
+   > **after** it in your profile. This is because zoxide wraps the existing
+   > `prompt` function to record directories. This is not a concern on other
+   > shells like Bash or Zsh.
 
    </details>
 
--- a/README.md
+++ b/README.md
@@ -281,6 +281,12 @@ zoxide can be installed in 4 easy steps:
    > ```powershell
    > Invoke-Expression (& { (zoxide init powershell | Out-String) })
    > ```
+   >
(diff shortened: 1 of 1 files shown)


LingyeNBird:
This issue has been bothering me for a while, and to prevent more people from falling into confusion, I believe there is a need for this note.
There may be a conflict in understanding here. zoxide has already stated that initialization should be done at the end of the config file, so why is an additional note necessary?
> That's because all related software states that the config file should be initialized at the end.
Therefore, either those theme configuration tools or zoxide, at least one of them needs to provide additional clarification.
After careful consideration, I believe it is reasonable to add a note here in zoxide.

If this is indeed not part of the plan, please help me close the PR.

PR #1260, opened 2026-07-06
Fix MSYS2 cygpath pwd substitution

### LINKED ISSUE #1259: [Bug]: Windows Git Bash (MSYS2): "not a directory: E:\builtin pwd -L" due to missing command substitution in __zoxide_pwd
### Describe the bug
When using `zoxide 0.10.0` in a Windows Git Bash (MSYS2) environment, running directory navigation commands like `cd ..` triggers the following error:

```text
zoxide: not a directory: E:\builtin pwd -L
```

This seems to happen specifically when navigating paths that reside on Windows drives (e.g., the `E:` drive, which is formatted as a Dev Drive in my case).

No such issue in `v0.9.x`

### Reproduction steps
1. Open Git Bash on Windows.
2. Install Zsh inside Git Bash.
3. Add the following to `.zshrc` and reload:
   ```bash
   eval "$(zoxide init zsh)"
   ```
4. Navigate to any directory on a Windows drive (e.g., `cd E:/some_folder`).
5. Observe the `not a directory: E:\builtin pwd -L` error.

### Environment
- **OS**: Windows 11
- **Shell**: Zsh (running inside Git Bash / MSYS2)
- **Zoxide Version**: 0.10.0

### Root cause analysis
The issue originates from the shell script generated by `zoxide init zsh` for Windows/MSYS2 environments. The generated `__zoxide_pwd` helper function incorrectly passes the literal string `"builtin pwd -L"` to `cygpath` instead of executing it.

**Generated code (incorrect):**
```bash
function __zoxide_pwd() {
    \command cygpath -w "builtin pwd -L" # Missing $(...)
}
```
Because of this, `cygpath` treats the string `"builtin pwd -L"` as a relative path and prepends the current drive letter (e.g., `E:\`). Zoxide then attempts to insert this invalid path (`E:\builtin pwd -L`) into its database, resulting in the "not a directory" error.

### Expected behavior
The `__zoxide_pwd` function should correctly capture the output of `builtin pwd -L` using command substitution.

**Expected generated code:**
```bash
function __zoxide_pwd() {
    \command cygpath -w "$(builtin pwd -L)"
}
```

### Current Workaround
I have temporarily patched this by overriding the function in my `.zshrc` immediately after the `eval` command:

```bash
eval "$(zoxide init zsh)"

# fix for 0.10.0
function __zoxide_pwd() {
    builtin pwd -L
}
```

*(Note: For better performance in MSYS2, completely skipping `cygpath` and just using `builtin pwd -L` might also be a valid option, as `cd` and zoxide's matching algorithms handle POSIX paths perfectly fine without converting them to Windows native paths on every call.)*

cyphercodes:
## Summary
- Pass the result of `pwd` to `cygpath` in Windows Bash/Zsh init snippets instead of passing the literal command string.
- Add a focused regression test for the Bash/Zsh template lines.

Fixes #1259

## Tests
- `cargo test windows_bash_and_zsh_pwd_substitute_before_cygpath`
- `cargo test`
- `cargo test --features nix-dev bash_bash`
- `cargo test --features nix-dev zsh_zsh`
- `cargo test --features nix-dev bash_shellcheck`
- `cargo test --features nix-dev zsh_shellcheck`
- `cargo fmt --check`
- `cargo clippy --all-features --all-targets -- -Dwarnings`
- `git diff --check`
- `cargo check --target x86_64-pc-windows-gnu`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Bash/Zsh: fix `z` failing on Cygwin/MSYS2 due to `cygpath` being passed a bad string.
+
 ## [0.10.0] - 2026-07-04
 
 ### Added
--- a/src/shell.rs
+++ b/src/shell.rs
@@ -33,6 +33,26 @@ make_template!(Tcsh, "tcsh.txt");
 make_template!(Xonsh, "xonsh.txt");
 make_template!(Zsh, "zsh.txt");
 
+#[cfg(test)]
+mod template_tests {
+    #[test]
+    fn windows_bash_and_zsh_pwd_substitute_before_cygpath() {
+        for (name, template) in [
+            ("bash", include_str!("../templates/bash.txt")),
+            ("zsh", include_str!("../templates/zsh.txt")),
+        ] {
+            assert!(
+                template.contains(r#"\command cygpath -w "$({{ pwd }})""#),
+                "{name} template should pass the output of pwd to cygpath"
+            );
+            assert!(
+                !template.contains(r#"\command cygpath -w "{{ pwd }}""#),
+                "{name} template should not pass the literal pwd command to cygpath"
+            );
+        }
+    }
+}
+
 #[cfg(feature = "nix-dev")]
 #[cfg(test)]
 mod tests {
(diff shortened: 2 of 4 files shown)


ajeetdsouza:
Thanks!

mojotx:
This bug is still affecting Windows users because 0.10.0 (July 4, 2026) was released three days before this fix merged, and it's still unreleased as of September 2026. I'm on 0.10.0 via WinGet on Windows 11 / Git Bash (MSYS2) and automatic directory tracking is broken — zoxide add manually works, but the prompt hook fails with zoxide: not a directory: C:\builtin pwd -L. Could we get a 0.10.1 patch release? 

Here's the workaround I'm using in my .bashrc in the meantime, in case someone else hits this bug and needs an immediate fix. 

```bash
if command -v zoxide >/dev/null 2>&1; then
    eval "$(zoxide init bash)"
    __zoxide_pwd() {
        command cygpath -w "$(builtin pwd -L)"
    }
fi
```

PR #1262, opened 2026-07-12
feat(macOS): add Finder alias path resolution

### LINKED ISSUE #855: Folder Alias not properly supported on macOS
CDing to an alias file in macOS correctly changes the current directory to the folder pointed to by the alias file.

zoxide only sees a file and emits an error.

DrewSkelton:
Adds Finder alias resolution for macOS.

Closes #855


ajeetdsouza:
While I'm okay with the premise of the PR, shelling out to osascript doesn't seem like a great solution. Are there better alternatives?

PR #1280, opened 2026-08-11
fix: reject non-finite ranks when importing

pcruz1905:
Importing a data file whose rank is `inf`, `nan`, or just larger than an
`f64` can hold throws away the rest of the database:

```console
$ cat ~/.z
/aaa|5|1000000000
/bbb|inf|1000000000
/ccc|3|1000000000
/ddd|10|1000000000

$ zoxide import z; echo "exit $?"
exit 0

$ zoxide query --list --score --all
   NaN /bbb
```

Three directories are gone and the command reports success.

`Database::age` scales every rank by `0.9 * max_age / total`. One non-finite
rank makes `total` infinite, so the factor is `0.0`: every other directory is
multiplied down to `0.0`, drops below the `1.0` threshold and is removed,
while the entry responsible survives, because `NaN < 1.0` is false like every
other comparison against it.

It does not stop there. The NaN keeps the total NaN, so `total_age > max_age`
is never true again and the database stops ageing permanently — 20 dirs at
rank 600 stay at a total of 12000 instead of being scaled back to 9000.

`last_accessed` is already validated, in that `parse::<u64>()` rejects
anything unreasonable and reports it with a line number. Doing the same for
the rank keeps the bad line out and leaves the rest of the file alone:

```console
$ zoxide import z
/home/user/.z:2: invalid entry: /bbb|inf|1000000000

$ zoxide query --list --score --all
   1.2 /aaa
   0.8 /ccc
```

`age` drops non-finite ranks too, so a database that has already been hit by
this recovers on the next run rather than staying stuck.

This covers `z`, `fasd`, `z.lua` and `zsh-z`, which share the parser.
`autojump` passes its rank through a sigmoid and was never affected.

#### Testing

`cargo test` (26 tests), `cargo +nightly fmt --check` and
`cargo clippy --all-targets -- -D warnings` are clean.

Added a parametrised parser test covering `inf`, `-inf`, `nan` and `1e400`
alongside the ranks that must keep working, including a path containing `|`,
and two `age` tests: one that the finite entries survive a poisoned database,
one that ageing still brings the total back under `max_age`. Reverting either
half of the fix on its own fails its tests and leaves the valid cases green.

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -121,6 +121,14 @@ impl Database {
     pub fn age(&mut self, max_age: Rank) {
         let mut dirty = false;
         self.with_dirs_mut(|dirs| {
+            // A non-finite rank would make the total non-finite too, which
+            // either scales every other rank to zero and drops it, or leaves
+            // the total NaN so that no comparison below ever holds and the
+            // database stops ageing entirely. Drop those entries instead.
+            let len_before = dirs.len();
+            dirs.retain(|dir| dir.rank.is_finite());
+            dirty |= dirs.len() != len_before;
+
             let total_age = dirs.iter().map(|dir| dir.rank).sum::<Rank>();
             if total_age > max_age {
                 let factor = 0.9 * max_age / total_age;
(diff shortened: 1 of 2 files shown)


xhon-pelushi (commented):
Checked this out and ran it. The whole suite passes (26/26), the `rsplitn` cases are well chosen — `/f|oo|3|100` covering a `|` inside the path is a nice touch — and the reasoning in both comments is accurate about why a poisoned total is worse than a single bad entry.

One thing though: the door is still open, because `is_finite()` on individual ranks does not imply a finite *total*.

### Finite ranks can still sum to infinity, and that still wipes the database

`f64` saturates on overflow rather than trapping, so two entries that each pass the new import check can still produce `total_age == inf`:

```rust
db.add_unchecked("/a",    1.0e308, now);   // finite -> survives retain()
db.add_unchecked("/b",    1.0e308, now);   // finite -> survives retain()
db.add_unchecked("/keep",   500.0, now);
db.age(10000.0);
```

```
before: len=3 total=inf total_finite=false
after age(): len=0 dirs=[]
```

Every entry is gone, `/keep` included. The path is the one your comment already describes: `total_age = inf`, so `factor = 0.9 * max_age / inf` is `0.0`, every `dir.rank *= factor` becomes `0.0`, and `if dir.rank < 1.0 { dirs.swap_remove(idx) }` removes all of them.

And `1e308` is admitted by the new import guard, since it is genuinely finite:

```
parse("1e308") finite=true
parse("1e400") finite=false   <- caught by your change
```

So a `.z` file with two `path|1e308|100` lines still costs the user their entire database after the next `age()`. Same consequence as the `inf` case, same input vector, just one exponent lower than the value you're rejecting.

### Suggestion

Guarding the total inside `age()` covers both shapes at a single choke point, and it does not depend on where the ranks came from:

```rust
let total_age = dirs.iter().map(|dir| dir.rank).sum::<Rank>();
if !total_age.is_finite() {
    // Overflowed or poisoned: scaling by a factor derived from this would
    // zero out every entry. Fall back to dropping only the outliers.
    ...
} else if total_age > max_age {
    ...
}
```

That would also make the import-side check defence-in-depth rather than the only thing standing between a corrupt `.z` file and an empty database — worth having, since `age()` is reachable from paths other than import, and `add_unchecked` is `pub`.

If you would rather keep this PR narrow, that is completely reasonable — the `inf`/`nan` rejection is a strict improvement on its own. Just flagging that the issue it is closing is not fully closed, so it does not get marked done and forgotten.

Tested on `c677bbc`, rustc 1.97.1, linux x86_64.

superbigcup325:
The follow-up suggested in the review — guarding the total inside `age()` — is now up as #1302, so this PR can stay narrow as proposed.

#1302 handles the case where every rank is finite but the sum still saturates (`1e308 + 1e308`), plus negative overflow; it also subsumes the `retain(is_finite())` guard added here, so the two compose cleanly: the import-side validation remains this PR's own contribution.

CC @pcruz1905 @xhon-pelushi

PR #1288, opened 2026-08-23
Switch to plaintext DB

ajeetdsouza:
Switches the database to an open format that works better with text editors and standard UNIX tools.


lengyijun:
Text-based history is easy to manipulate, so should we remove some subcommands?

ajeetdsouza:
Yeap. We can remove `zoxide edit` after this, it was always a bit clunky. We can either remove the subcommand entirely, or make it just open up the database file in `$EDITOR`.

PR #1300, opened 2026-09-07
import: stop iterating after input read errors

cuishuang:
Stop import iterators after an input read error so persistent I/O failures are reported once instead of being emitted indefinitely.

This applies to the z, autojump, and Atuin importers. Importers that reuse z::Iter, including fasd, z.lua, and zsh-z, are covered automatically.


ajeetdsouza:
Do you have an example of why this is a problem? The importer has been deliberately written to show multiple errors, in case a user had multiple invalid lines.

cuishuang:
> Do you have an example of why this is a problem? The importer has been deliberately written to show multiple errors, in case a user had multiple invalid lines.

Thanks. I agree that multiple malformed lines should all be reported, and this change preserves that behavior.

The case addressed here is different: `done` is only set when `read_until()` itself returns an I/O error. Errors returned by `parse_line()` still produce an item-level error and continue to the next record as before.

For example, a database stored on an NFS/FUSE filesystem may start returning a persistent `EIO` after the file has been opened. A deterministic equivalent is the `BrokenReader` in the tests, whose `fill_buf()` always returns an error.

With the current implementation, that produces:

test:1: boom
test:2: boom
test:3: boom
...

indefinitely, because the import loop calls `next()` again after every error, while the reader is not required to make progress after an error. The command never reaches `db.save()` and must be interrupted manually.

`BufRead::read_until()` already retries `ErrorKind::Interrupted`. For other errors, it returns the error without guaranteeing that a subsequent call will advance the stream:
https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_until

After this change, malformed rows are still all reported, while a source-level read failure is reported once and terminates that iterator.

